PR #129, opened 2022-12-03
Add a model package and a fetch component

AnonymusRaccoon:


AnonymusRaccoon:
Test fails on next only pages.

PR #158, opened 2023-03-20
Add ApiKeys to the API

zoriya:
Also, delete the weird configurations settings that still exists in the back.
Cleanup env variables for already existing confugrations
Make real error messages for authentication (be it ApiKeys or JWT)


zoriya:
Still missing:
 - Making requests without any apikeys or jwt now works
 - Making a request with the good api key fail.

PR #262, opened 2023-12-09
Front: Minor UI tweaks

Arthi-chaud:
# 'Recommended' Component

## Border Radius of Poster

Before:
Screenshot_2023-12-08_at_14 24 561

After:
Screenshot_2023-12-08_at_14 24 49

## Fix Play button Position
Before:
Screenshot 2023-12-09 at 08 46 25

After:

Screenshot 2023-12-09 at 08 50 59

## Horizontal Padding
Before:
Screenshot 2023-12-09 at 08 54 27
After:
Screenshot 2023-12-09 at 08 54 53

## Poster Text Color
Before: 
Screenshot 2023-12-09 at 08 58 14
Screenshot 2023-12-09 at 08 58 23

After: 
Screenshot 2023-12-09 at 08 58 28
Screenshot 2023-12-09 at 08 58 37

# Poster Border

Before: 
Screenshot 2023-12-08 at 18 30 10
After:
Screenshot 2023-12-08 at 18 30 18

# Item: Name on 2 lines

Before:
Screenshot 2023-12-09 at 08 44 27
After:
Screenshot 2023-12-09 at 08 44 19


front/packages/primitives/src/constants/image-border-radius.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Kyoo is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Kyoo. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+export const ImageBorderRadius = 10;
```
zoriya, line 21: I think we should define all constants in a single files, I don't want to end up with 30 files only containing an int.

PS: I think, not sure, that PascaleCased variables are usually reserved for react component

front/packages/ui/src/browse/grid.tsx
```diff
@@ -157,7 +157,10 @@ export const ItemGrid = ({
 			</PosterBackground>
 			<Skeleton>
 				{isLoading || (
-					<P numberOfLines={1} {...css([{ marginY: 0, textAlign: "center" }, "title"])}>
+					<P
+						numberOfLines={subtitle ? 1 : 2}
```
zoriya, line 161: If I recall correctly, this is only used for collections. Is this intended?

PS: I was thinking of adding something like `3 items` as a subtitle for collection, so that would cancel that.

front/packages/ui/src/home/recommanded.tsx
```diff
@@ -89,7 +90,7 @@ export const ItemDetails = ({
 					height: ItemDetails.layout.size,
 					flexDirection: "row",
 					bg: (theme) => theme.variant.background,
-					borderRadius: 6,
+					borderRadius: 12, // to make the border stick to the poster
```
zoriya, line 93: I don't understand why you are not using the constant for image radius. Is it because you need to add borderWidth to borderRadius ? (if that's the case, why css, why).

If that's the case, you can use `calc(px(ImageRadiusConst), "+", ts(.25))` using yoshiki/native

zoriya: Or you could just set it to `borderRadius: ImageRadiusConst` and set the poster's style to `borderRadius: 0` instead, you already set borderRight to 0 anyways.

front/packages/ui/src/home/recommanded.tsx
```diff
@@ -125,7 +127,7 @@ export const ItemDetails = ({
 					})}
 				>
 					<Skeleton {...css({ width: percent(100) })}>
-						{isLoading || <P {...css([{ m: 0 }, "title"])}>{name}</P>}
+						{isLoading || <P {...css([{ m: 0, color: "white" }, "title"])}>{name}</P>}
```
zoriya, line 130: you should use theme.colors.white instead of a hardcoded white. Kyoo never uses #ffffff white.

front/packages/ui/src/home/recommanded.tsx
```diff
@@ -152,19 +154,20 @@ export const ItemDetails = ({
 					{...css({
 						bg: (theme) => theme.themeOverlay,
 						flexDirection: "row",
+						pX: 4,
 						justifyContent: "space-between",
 						minHeight: px(50),
 					})}
 				>
-					{(isLoading || genres) && (
```
zoriya, line 159: doesn't that result in skeletons being always shown for collections?

zoriya (approved):
Looks way better, left a few questions/nit picks

front/packages/ui/src/home/recommanded.tsx
Arthi-chaud: Nope :), this is what the change two lines below is for

front/packages/ui/src/browse/grid.tsx
Arthi-chaud: For now, yes, this would be mainly applicable to collections.

However, would it be possible for a show not to have a year? This would be another 'use case'.
Let me know what you think

front/packages/ui/src/home/recommanded.tsx
zoriya: Ok, I just saw that, but you disabled skeletons for everything. `Array(3).map(() => 2)` result in an empty array. Array(3) create an empty array with a capacity of 3 so .map() returns an empty array (that's why you need [...Array(3)] instead.

front/packages/ui/src/browse/grid.tsx
zoriya: True, it is possible, but I have never seen it happens. If this was intended, that's good.

front/packages/ui/src/home/recommanded.tsx
Arthi-chaud: Solved by using 'flex-end' on the parent component

PR #271, opened 2023-12-18
Front: Rework Episode Tile to allow Expansion of its description

Arthi-chaud:
Before:
Screenshot 2023-12-20 at 16 20 27

After:
Screenshot 2023-12-20 at 16 20 56

### DIFF
--- a/front/packages/ui/src/details/episode.tsx
+++ b/front/packages/ui/src/details/episode.tsx
@@ -21,6 +21,7 @@
 import {
 	focusReset,
 	H6,
+	IconButton,
 	ImageBackground,
 	ImageProps,
 	important,
@@ -28,6 +29,7 @@ import {
 	P,
 	Skeleton,
 	SubP,
+	tooltip,
 	ts,
 } from "@kyoo/primitives";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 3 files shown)


front/packages/ui/src/details/episode.tsx
```diff
@@ -99,7 +103,10 @@ export const EpisodeBox = ({
 							borderStyle: "solid",
 						},
 						more: {
-							display: "none",
+							opacity: 0,
```
zoriya, line 106: I don't think this is used, is it?

front/packages/ui/src/details/episode.tsx
```diff
@@ -293,32 +301,56 @@ export const EpisodeLine = ({
...
-					)}
+					<Skeleton>
+						{isLoading || <P numberOfLines={descriptionExpanded ? undefined : 3}>{overview}</P>}
+					</Skeleton>
+					<IconButton
+						{...css(["more"])}
+						icon={descriptionExpanded ? ExpandLess : ExpandMore}
+						{...tooltip(t(descriptionExpanded ? "misc.collapse" : "misc.expand"))}
+						onPress={(e) => {
+							e.preventDefault();
```
zoriya, line 350: Is the preventDefault() needed?

zoriya (approved):
Thanks for tackling this!

front/packages/ui/src/details/episode.tsx
Arthi-chaud: Yes it is: the `more` class now sets the opacity of an element to 1 (instead of setting `display` to none) when the parent is hovered.
However, the `less` class is not used. Removing it right away.

front/packages/ui/src/details/episode.tsx
Arthi-chaud: Yes it is. It prevents the click event from being passed to the parent `<A>`

PR #277, opened 2023-12-21
Allow download quality to be selected in settings

zoriya:
Missing:
 - Download items before transcoding as finished (something like https://stackoverflow.com/questions/59757312/how-do-i-create-a-stream-for-actix-web-httpresponse-to-send-a-file-chunk-by-chun/64486497#64486497)
 - Clear the download cache


zoriya:
Closed in favor of #290

PR #290, opened 2024-01-19
Add transcoded download support

zoriya:
The length of the stream is invalid (since ffmpeg seek at the beginning of the file to set the correct headers for length, but we already sent this part of the file to the client) so this makes bugged streams on android and players can't display a valid progress bar.
This could be fixed by using another container but only mkv allow all fonts/attachment so...

The other issue is that we don't know how big the file will be so the download can't display a progress bar (on android the current implementation always display -1%)

I wonder if we could just transcode the file completly and then send it.


zoriya:
Potential solution: download the file in a pseudo fs (like a worker) and after this is done, download the file from the worker via the browser (since the file would then be local this is instant and does not require user consent). This is kinda what mega does

PR #295, opened 2024-01-21
Front: Mediainfo Popup

### LINKED ISSUE #287: Mediainfo popup
### Feature description

Episodes and movies should have a button to display their mediainfo. There is already a route (/api/video/{episode/movie}/{slug}/info) in the transcoder for that.

- [ ] The mediainfo popup should be accessible from movies' and episodes' context menus
  - [ ] Not shows
- [ ] The mediainfo popup should also be accessible from the player

Arthi-chaud:
Media Info Popup for Episodes and Movies. (closes #287)
Screenshot 2024-01-21 at 13 10 54

A few questions/problems:

- [x] Which Icon should we use for context menus? (Currently we use `Info`, but it is the same as `Go To Show`)
- [ ] Whether it is from a context menu or a movie page's header, we always need to click the button twice for the popup to work


front/packages/models/src/resources/watch-info.ts
```diff
@@ -29,9 +29,42 @@ const getDisplayName = (sub: Track) => {
...
+ */
+export const VideoP = z.object({
+	/**
+	 * The Codec of the Video Track.
+	 * E.g. "AVC"
+	 */
+	codec: z.string(),
+	/**
+	 * The Quality of the Video
+	 * E.g. "1080p" (TODO: FIND ACTUAL ENUM)
```
zoriya, line 46: The enum is actually not typed in the front, I added it inline here https://github.com/zoriya/Kyoo/blob/cc9b6e69c83193abbad6332cca08a9f510d84938/front/packages/models/src/resources/user.ts#L42 (not on master), it could be nice to extract it into another const.

front/packages/ui/src/components/media-info.tsx
```diff
@@ -0,0 +1,125 @@
...
+			},
+			{
+				[t("mediainfo.video")]: video
+					? `${video.width}x${video.height} (${video.quality}) - ${formatBitrate(
+							video.bitrate,
+					  )} - ${video.codec}`
+					: undefined,
+			},
+			audios === undefined
+				? { [t("mediainfo.audio")]: undefined }
```
zoriya, line 72: I did not test it yet but I think that you always show the skeleton if an item does not have audio or subtitles.

zoriya: Never mind, you filter out undefined just below

front/packages/ui/src/components/media-info.tsx
```diff
@@ -0,0 +1,125 @@
...
+	mediaType: "episode" | "movie";
+	mediaSlug: string;
+}) => {
+	const mediaInfoQuery = {
+		path: ["video", mediaType, mediaSlug, "info"],
+		parser: WatchInfoP,
+	};
+	return (
+		<Popup>
+			{({ css }) => (
```
zoriya, line 115: You can just useYoshiki() inside the MediaInfoTable, you don't need to pass it from here. Popup gives one because it overrides the styles for children but since MediaInfoTable is already a child of Popup, calling useYoshiki() will give the correct theme.

front/packages/ui/src/components/media-info.tsx
```diff
@@ -0,0 +1,125 @@
...
+		if (trackTable.length == 0) {
+			return undefined;
+		}
+		return trackTable.reduce(
+			(collected, audioTrack, index) => ({
+				...collected,
+				// If there is only one track, we do not need to show an index
+				[trackTable.length == 1 ? t(s) : `${t(s)} ${index + 1}`]: [
+					audioTrack.displayName,
+					audioTrack.isDefault ? t("Default") : undefined,
```
zoriya, line 48: Default and Forced are not defined in en.json/fr.json

zoriya:
What about this icon https://fonts.google.com/icons?selected=Material+Symbols+Outlined:movie_info:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=info ?

zoriya:
For the double press needed, I don't understand why but it works with:

```tsx
	const [isInfoOpen, setInfoOpen] = useState(false);
	return <>
			<Menu>
				<Menu.Item onSelect={() => setInfoOpen(true)} />
			</Menu>
			{isInfoOpen && type !== "show" && (
				<MediaInfoPopup mediaType={type} mediaSlug={slug} close={() => setInfoOpen(false)} />
			)}
	</>
```

zoriya:
I think it could be nice to also add duration and file size to this popup. File size is currently not available, but I'll think about adding it to the endpoint. Duration is available in a field named length

zoriya:
On android, the popup can overflow and the user is not able to scroll it. You probably need to wrap it with a scrollview.

front/packages/ui/src/components/media-info.tsx
```diff
@@ -0,0 +1,126 @@
...
+				? { [t("mediainfo.subtitles")]: undefined }
+				: formatTrackTable(subtitles, "mediainfo.subtitles"),
+		] as const
+	).filter((x) => x !== undefined) as Record<string, string | undefined>[];
+	return (
+		<View>
+			{table.map((g) =>
+				Object.entries(g).map(([label, value], index, l) => (
+					<>
+						<View key={`media-info-${label}`} {...css({ flexDirection: "row" })}>
```
zoriya, line 85: This key is useless since the component is wrapped on a fragment. The key should directly be appiled to the fragment.

(this can be done by using the verbose `<>` like so `<Fragment key={}>`

PR #388, opened 2024-04-07
back: migrations: fix postgres config

Intuinewin:
Hello,

The postgres server hostname and port are currently hardcoded in the migration dockerfile. Small fix to use the existing environment variables


zoriya:
Thanks!

zoriya (approved):

PR #423, opened 2024-04-16
Add Simplified Chinese translation

tardlk:
Add Simplified Chinese translation


zoriya (approved):

zoriya:
Thanks!

zoriya:
NOTE: We should definitely look into making translations easier to create. Adding translations to js objects feels wrong. Maybe check localize out? Or something like that?

PR #447, opened 2024-04-28
scanner: supports tmdb episode groups with multiple groups

Intuinewin:
Sometimes episodes groups in tmdb have multiple groups inside them, kyoo should use all of them instead of the first one.

Example episode group: https://www.themoviedb.org/tv/203819-d-argent-et-de-sang/episode_group/65a19141f04d010131792b95


zoriya:
Good catch, did not thought absolute groups could have subgroups.

Issue #466, opened 2024-05-02
Improved anime detection (Like `Attack on Titan Season 3 Part 2 12.mkv`)


zoriya:
### Feature description

Some attempts were made and reported in #352, but nothing good yet. Putting this here for better tracking.

I'm not sure how we could implement that. Something like https://github.com/Anime-Lists/anime-lists seems like the only possible solution, but they only support tvdb (and their API is premium).

Movies that are direct sequels of series should also be handled with something like that (for example, Made In Abyss's 3rd movie is a sequel to the first season and the prequel to the second season). This would also help for episodes marked in the season 0 on TMDB (like SNK's last episodes)

Almightily:
Good day.
I have an idea of how to resolve this, perhaps we can somehow manually mark episodes and connect them to the episodes on the thmdb, in the admin panel, or file name for example.

zoriya:
I want to avoid any manual work, especially for scanning items. This weird naming scheme is pretty common on animes, so users would need to match them quite frequently. We should really have built-in support for those.

zoriya:
Cases like `Jujutsu Kaisen S02` (marked as part of S1 on TMDB) should also be handled by this. (this case was reported on #455)

zoriya:
This was discussed on #352, but the current idea is:

- Search offline using [Anime-Lists/anime-lists](https://github.com/Anime-Lists/anime-lists/).
- If we get a match:
	- Get metadata from TVDB using id given in `anime-lists`.
	- Complete metadata with TheMovieDB (some data only exists there and quality is often better)
- If we don't get a match:
	- Search on TVDB or TheMovieDB manually. Still not sure which one to pick. 
- Add other ids/link of `anime-lists` to the saved `externalIds` of the episodes/movies/shows

PR #494, opened 2024-05-14
Basic tvdb provider

zoriya:
This implements a basic TVDB provider for #466
This provider does not implement movies since the anime list uses TMDB for movies so it won't be needed.

Still has lots of bugs, need iterations.

(based on #493)


zoriya:
Note: TVDB does not return placeholder season/episode titles (like `Season 2` or `Episode 2`), we should add fallback like this in the front.

Right now, not having season title is not handled in the front and the component stays in loading state.

This should be fixed in a follow up PR.

PR #502, opened 2024-05-19
Manually fix translations and update node packages

zoriya:


zoriya:
Still need to test on android, waiting for CI to build.

PR #528, opened 2024-06-08
Add media type filter

scme0:
# Background

The first step to adding a series of filters on the browse page.

Discord discussion: https://discord.com/channels/1216460898139635753/1241683417180143646/1241683417180143646

# Results

I've been working on a searchable multi-select drop down component but it's taking a bit longer than I had hoped as I haven't had a lot of time in the last few weeks. I feel like this would be the ideal way to filter things to genres where there are a lot of options and you might want to select multiple.

While working on that, and to get things started in this area, I thought we could add this basic Media Type filter which will allow users to select the media type they wish to browse. I thought the `Menu` component was suitable as there are only 3 options.

# Demo

Screen Recording 2024-06-08 at 18 26 43

### DIFF
--- a/front/packages/ui/src/browse/header.tsx
+++ b/front/packages/ui/src/browse/header.tsx
@@ -109,8 +109,6 @@ export const BrowseSettings = ({
 }) => {
 	const { css, theme } = useYoshiki();
 	const { t } = useTranslation();
-	const filters: string[] = [];
-	// TODO: implement filters in the front.
 
 	return (
 		<>
@@ -177,26 +175,6 @@ export const BrowseSettings = ({
 						))}
 					</Menu>
 				</View>
-			</View>
-			<View
-				{...css({
-					flexDirection: "row-reverse",
-					alignItems: "center",
-					marginX: ts(4),
-					marginY: ts(1),
-					zIndex: 1,
-				})}
-			>
-				{filters.length !== 0 && (
-					<View {...css({ flexGrow: 1, flexDirection: "row", alignItems: "center" })}>
-						{/*<Icon icon={Style} {...css({ marginX: ts(1) })} />*/}
-						{filters.map((x) => (
-							<div style={{paddingRight: ".25rem"}}>
-								<Chip key={x} label={x} size={"small"} />
-							</div>
-						))}
-					</View>
-				)}
 			</View>
 		</>
 	);
--- a/front/packages/ui/src/browse/header.tsx
+++ b/front/packages/ui/src/browse/header.tsx
@@ -56,13 +56,14 @@ const MediaTypeTrigger = forwardRef<View, PressableProps & { mediaType: MediaTyp
 		const { t } = useTranslation();
 		const labelKey =
 			mediaType !== MediaTypeAll ? `browse.mediatypekey.${mediaType.key}` : "browse.mediatypelabel";
+		const icon = mediaType !== MediaTypeAll ? mediaType?.icon ?? FilterList : FilterList;
 		return (
 			<PressableFeedback
 				ref={ref}
 				{...css({ flexDirection: "row", alignItems: "center" }, props as any)}
 				{...tooltip(t("browse.mediatype-tt"))}
 			>
-				<Icon icon={mediaType?.icon ?? FilterList} {...css({ paddingX: ts(0.5) })} />
+				<Icon icon={icon} {...css({ paddingX: ts(0.5) })} />
 				<P>{t(labelKey as any)}</P>
 			</PressableFeedback>
 		);
--- a/front/packages/ui/src/browse/header.tsx
+++ b/front/packages/ui/src/browse/header.tsx
@@ -18,27 +18,18 @@
  * along with Kyoo. If not, see <https://www.gnu.org/licenses/>.
  */
 
-import {
-	HR,
-	Icon,
-	IconButton,
-	Menu,
-	P,
-	PressableFeedback,
-	tooltip,
-	ts,
-} from "@kyoo/primitives";
+import { HR, Icon, IconButton, Menu, P, PressableFeedback, tooltip, ts } from "@kyoo/primitives";
 import ArrowDownward from "@material-symbols/svg-400/rounded/arrow_downward.svg";
 import ArrowUpward from "@material-symbols/svg-400/rounded/arrow_upward.svg";
 import GridView from "@material-symbols/svg-400/rounded/grid_view.svg";
 import Sort from "@material-symbols/svg-400/rounded/sort.svg";
 import FilterList from "@material-symbols/svg-400/rounded/filter_list.svg";
 import ViewList from "@material-symbols/svg-400/rounded/view_list.svg";
-import {forwardRef} from "react";
+import { forwardRef } from "react";
 import { useTranslation } from "react-i18next";
 import { type PressableProps, View } from "react-native";
 import { useYoshiki } from "yoshiki/native";
-import {Layout, MediaType, MediaTypeAll, SearchSort, SortOrd} from "./types";
+import { Layout, MediaType, MediaTypeAll, SearchSort, SortOrd } from "./types";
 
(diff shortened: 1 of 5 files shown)


front/packages/ui/src/browse/header.tsx
```diff
@@ -81,58 +113,91 @@ export const BrowseSettings = ({
...
+				{filters.length !== 0 && (
+					<View {...css({ flexGrow: 1, flexDirection: "row", alignItems: "center" })}>
+						{/*<Icon icon={Style} {...css({ marginX: ts(1) })} />*/}
+						{filters.map((x) => (
+							<div style={{paddingRight: ".25rem"}}>
+								<Chip key={x} label={x} size={"small"} />
+							</div>
+						))}
+					</View>
+				)}
```
zoriya, line 199: This is never used, it was a WIP filter implementation. I think it can be removed with the new implementation.

front/packages/ui/src/browse/index.tsx
```diff
@@ -52,27 +52,30 @@ export const itemMap = (
...
 });
 
-const query = (sortKey?: SortBy, sortOrd?: SortOrd): QueryIdentifier<LibraryItem> => ({
+const query = (sortKey?: SortBy, sortOrd?: SortOrd, mediaTypeKey?: string): QueryIdentifier<LibraryItem> => ({
 	parser: LibraryItemP,
 	path: ["items"],
 	infinite: true,
 	params: {
 		sortBy: sortKey ? `${sortKey}:${sortOrd ?? "asc"}` : "name:asc",
+		filter: mediaTypeKey && mediaTypeKey !== "none" ? `kind eq ${mediaTypeKey}` : undefined,
```
zoriya, line 61: Why are you checking for `none` here? I dont think it is used anywhere elsee

front/packages/ui/src/browse/header.tsx
```diff
@@ -81,58 +113,91 @@ export const BrowseSettings = ({
...
+					<Menu Trigger={MediaTypeTrigger} mediaType={mediaType}>
+						{availableMediaTypes.map((x) => (
+							<Menu.Item
+								key={x.key}
+								label={t(`browse.mediatypekey.${x.key}` as any)}
+								selected={mediaType === x}
+								icon={x.icon}
+								onSelect={() => {
+									if (mediaType === x || x === AllMediaTypes) {
+										setMediaType(undefined)
```
zoriya, line 171: Is there a reason you use undefined here instead of all?

front/packages/ui/src/browse/index.tsx
```diff
@@ -52,27 +52,30 @@ export const itemMap = (
 		item.kind === "show" ? item.watchStatus?.unseenEpisodesCount ?? item.episodesCount! : null,
 });
 
-const query = (sortKey?: SortBy, sortOrd?: SortOrd): QueryIdentifier<LibraryItem> => ({
+const query = (sortKey?: SortBy, sortOrd?: SortOrd, mediaTypeKey?: string): QueryIdentifier<LibraryItem> => ({
```
zoriya, line 55: Is there a reason you use strings instead of enum values?

zoriya:
Really good PR, with a nice usage of components and existing libraries!

I added some nitpicks here and there but it looks really good!

I'll test this tonight or tomorrow!

front/packages/ui/src/browse/header.tsx
scme0: I was planning to introduce the filter chips once we have other filters (like genre etc) but you're right it shouldn't be in this PR. I'll remove it for now and we can add it back in if works well with other filters.

front/packages/ui/src/browse/index.tsx
scme0: Good point, I've cleaned this up so it uses an enum, but also I've adjusted it so that once we parse the `mediaTypeParam` string, we get a `MediaType` object which is never undefined. I think this makes it easier to reason about and simplifies the code.

front/packages/ui/src/browse/header.tsx
scme0: I've fixed this now so undefined is not valid 👍

front/packages/ui/src/browse/index.tsx
scme0: Fixed 👍

scme0:
I've just realised that the search query doesn't support filters yet. I guess we'll need to implement that before we can bring this in otherwise the UX is confusing, unless we hide the filters on the browse page when someone does a manual search.

scme0:
I've created another PR to add filtering to the search endpoint in the backend: https://github.com/zoriya/Kyoo/pull/530.

Still a WIP as I haven't tested it yet.

front/packages/ui/src/browse/header.tsx
```diff
@@ -60,79 +50,110 @@ const SortTrigger = forwardRef<View, PressableProps & { sortKey: string }>(funct
...
+			mediaType !== MediaTypeAll ? `browse.mediatypekey.${mediaType.key}` : "browse.mediatypelabel";
+		const icon = mediaType !== MediaTypeAll ? mediaType?.icon ?? FilterList : FilterList;
+		return (
+			<PressableFeedback
+				ref={ref}
+				{...css({ flexDirection: "row", alignItems: "center" }, props as any)}
+				{...tooltip(t("browse.mediatype-tt"))}
+			>
+				<Icon icon={icon} {...css({ paddingX: ts(0.5) })} />
+				<P>{t(labelKey as any)}</P>
```
zoriya, line 67: I don't think you need the `as any` anymore. Typescript should be able to infer labelKey as a valid translation key (if not please dismiss this comment)

front/packages/ui/src/browse/header.tsx
```diff
@@ -60,79 +50,110 @@ const SortTrigger = forwardRef<View, PressableProps & { sortKey: string }>(funct
...
-		<View
-			{...css({
-				flexDirection: "row-reverse",
-				alignItems: "center",
-				marginX: ts(4),
-				marginY: ts(1),
-			})}
-		>
-			{filters.length !== 0 && (
+		<>
```
zoriya, line 98: This fragment is now usless and can be removed

front/packages/ui/src/search/index.tsx
```diff
@@ -52,15 +54,18 @@ export const SearchPage: QueryPage<{ q?: string }> = ({ q }) => {
 	const pageStyle = usePageStyle();
 	const { t } = useTranslation();
 	const [sort, setSort] = useParam("sortBy");
+	const [mediaTypeParam, setMediaTypeParam] = useParam("mediaType");
 	const sortKey = (sort?.split(":")[0] as SearchSort) || SearchSort.Relevance;
 	const sortOrd = (sort?.split(":")[1] as SortOrd) || SortOrd.Asc;
 	const [layout, setLayout] = useState(Layout.Grid);
 
+	const mediaType = getMediaTypeFromParam(mediaTypeParam);
+	console.log("search: mediaType", mediaType);
```
zoriya, line 63: don't forget to remove this!

front/packages/ui/src/browse/header.tsx
scme0: Unfortunately it still seems required. I'm not sure if it can infer the type because I'm using string interpolation. I'm not sure if there is a way to make typescript play ball. 🤔

front/packages/ui/src/browse/header.tsx
scme0: Good catch, thanks.

front/packages/ui/src/browse/header.tsx
zoriya: I thought `mediaType.key` was known to be a string literal union and typescript would be smart enough. Too bad, let's keep the `as any`

zoriya (approved):
Perfect! Will wait to merge #530 before this, but this seems ready for merge!

zoriya:
Thanks for your hard work on this! Really happy to have this now!

PR #529, opened 2024-06-09
Add lint:fix script to auto-apply linting fixes

scme0:
I noticed that this seemed to be missing so I added it in quickly.


zoriya:
Thanks!

PR #530, opened 2024-06-09
Add filter to meilisearch in backend

scme0:
Required if we want to seamlessly add filtering to the browse page. Otherwise filtering won't work when users use the Search bar.

I've now tested this and it seems to work with enums (`genres`) and dates (`airDate`, `startAir`, `endAir`).

### DIFF
--- a/back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
+++ b/back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
@@ -39,7 +39,7 @@ internal static class FilterExtensionMethods
 		return value switch
 		{
 			null => null,
-			string s => s.Any(char.IsWhiteSpace) ? $"\"{s}\"" : s,
+			string s => s.Any(char.IsWhiteSpace) ? $"\\\"{s}\\\"" : s,
 			DateTimeOffset dateTime => dateTime.ToUnixTimeSeconds(),
 			DateOnly date => date.ToUnixTimeSeconds(),
 			_ => value
--- a/back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
+++ b/back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
@@ -1,10 +1,11 @@
+using System.Collections;
 using System.ComponentModel.DataAnnotations;
 using Kyoo.Abstractions.Models.Utils;
 using static System.Text.Json.JsonNamingPolicy;
 
 namespace Kyoo.Meiliseach;
 
-public static class FilterExtensionMethods
+internal static class FilterExtensionMethods
 {
 	public static string? CreateMeilisearchFilter<T>(this Filter<T>? filter)
 	{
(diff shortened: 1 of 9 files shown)


zoriya:
Tested and it work for simple cases, what does not work is mostly due to our current implementation of meilisearch syncing.

Enums are stored as integer and not string values so comparisons fail
DateTime/DateOnly are stored as string (and not unix timestamps) so meilisearch treat them as basic strings (see [doc](https://www.meilisearch.com/docs/learn/fine_tuning_results/working_with_dates))

back/src/Kyoo.Meilisearch/SearchManager.cs
```diff
@@ -99,11 +99,19 @@ public SearchManager(MeilisearchClient client, ILibraryManager libraryManager)
...
+			"items",
+			query,
+			filter.CreateMeilisearchFilter(),
+			sortBy,
+			pagination,
+			include
+		);
 	}
 
 	/// <inheritdoc/>
```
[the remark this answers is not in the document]
scme0: Done 👍

back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
```diff
@@ -0,0 +1,68 @@
...
+	{
+		return $"{CamelCase.ConvertName(property)} {@operator} {value.InMeilisearchFormat()}";
+	}
+
+	private static object? InMeilsearchFilterFormat(this object? value)
+	{
+		return value switch
+		{
+			null => null,
+			string s => s.Any(char.IsWhiteSpace) ? $"\"{s}\"" : s,
```
zoriya, line 42: String containing `"` would fail there, I think they should be escaped (I have not checked the meili documentation for how to escape).

back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+	public static string? CreateMeilisearchFilter<T>(this Filter<T>? filter)
+	{
+		return filter switch
+		{
+			Filter<T>.And and
+				=> $"({and.First.CreateMeilisearchFilter()}) AND ({and.Second.CreateMeilisearchFilter()})",
+			Filter<T>.Or or
+				=> $"({or.First.CreateMeilisearchFilter()}) OR ({or.Second.CreateMeilisearchFilter()})",
+			Filter<T>.Gt gt => CreateBasicFilterString(gt.Property, ">", gt.Value),
```
zoriya, line 18: You forgot to call `.InMeiliseabhFilterFormat()` for all values here, no?

zoriya: mb, it's on the function below

back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
```diff
@@ -0,0 +1,68 @@
...
+		{
+			null => null,
+			string s => s.Any(char.IsWhiteSpace) ? $"\"{s}\"" : s,
+			DateTimeOffset dateTime => dateTime.ToUnixTimeSeconds(),
+			DateOnly date => date.ToUnixTimeSeconds(),
+			_ => value
+		};
+	}
+
+	public static object? InMeilisearchFormat(this object? value)
```
zoriya, line 49: I think this should be defined in the `MiliSync` file. Having public expansion methods on objects always feels wrong (plus it would help with location of behaviors).
I'd rather have `ToUnixTimeSeconds` public.

back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
scme0: I was following the documentation where it says to wrap strings with spaces in them with either double or single quotes: https://www.meilisearch.com/docs/learn/fine_tuning_results/filtering#examples

But, to be fair I haven't had a chance to test this because the number of properties I can filter on are limited to `kind, status, airDate, startAir, endAir, studioId, genres` so we could just remove this handing and add it if/when it's required.

back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
scme0: Yep that's fair, I'll move it 👍

back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
scme0: Sorry, I didn't read your comment correctly the first time. Yes they might need to be escaped. I might just remove this for now. We can add it if/when we have properties we can filter on that have spaces in the text.

zoriya: I'd rather leave it on, the API would already work with every filter that way.

I just checked in their issue tracker, we can escape `"` with `\`.

scme0:
1. No problems, that makes sense.
2. Ah yeah sorry. I was only thinking about filtering (which is restricted to certain properties) but of course you can do a search over all the properties with the query parameter. 😅 
3. Yeah this a tricky one. You could potentially trigger it as part of a database migration? It would probably take a while though as you'd need to drop the meilisearch database and then iterate every document type and load them back in. But if it's not done, searches wont work correctly for some users. Maybe as an escape hatch we could add a button a bit like "rescan library" which reloads data into meilisearch.

back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
scme0: Ok I've added escape chars 👌

back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
scme0: Done! 🚀

scme0:
Ah yep, that would probably work. Meilisearch should be available when the migration is running as it doesn't depend on anything else?

I could look into it later.

How often does the metadata refresh?

scme0:
@zoriya This is ready for another review when you have some time 🙏

zoriya:
Apart from the escaping issue, it looks good. Do you want me to add the meilisearch's migration, or do I leave that to you?

back/src/Kyoo.Meilisearch/FilterExtensionMethods.cs
```diff
@@ -0,0 +1,52 @@
...
+	{
+		return $"{CamelCase.ConvertName(property)} {@operator} {value.InMeilsearchFilterFormat()}";
+	}
+
+	private static object? InMeilsearchFilterFormat(this object? value)
+	{
+		return value switch
+		{
+			null => null,
+			string s => s.Any(char.IsWhiteSpace) ? $"\\\"{s}\\\"" : s,
```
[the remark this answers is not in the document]
scme0: Ah yep, I understand what you meant before now. Sorry! I'll fix it up.

scme0:
> Apart from the escaping issue, it looks good. Do you want me to add the meilisearch's migration, or do I leave that to you?

What migration are you talking about? Adding a database migration which updates all documents in meilisearch?

zoriya:
Yes

zoriya (approved):

PR #539, opened 2024-06-16
WIP add migration to reindex meilisearch

scme0:


zoriya:
Finally, I simply made meilisearch sync with postgres on every boot of the backend (can be disabled with an env var for replicas, for example). This will be more resilient and way simpler.

PR #541, opened 2024-06-16
Transcoder: Support audio-only files

Arthi-chaud:
Not really useful *for* Kyoo, but adding support for audio-only files in Transcoder

### DIFF
--- a/transcoder/src/keyframes.go
+++ b/transcoder/src/keyframes.go
@@ -2,7 +2,6 @@ package src
 
 import (
 	"bufio"
-	"errors"
 	"fmt"
 	"log"
 	"os/exec"
@@ -27,9 +26,6 @@ type KeyframeInfo struct {
 func (kf *Keyframe) Get(idx int32) float64 {
 	kf.info.mutex.RLock()
 	defer kf.info.mutex.RUnlock()
-	if len(kf.Keyframes) == 0 {
-		return float64(idx * 2)
-	}
 	return kf.Keyframes[idx]
 }
 
@@ -84,7 +80,7 @@ func GetKeyframes(sha string, path string) *Keyframe {
 				return
 			}
 
-			err := getKeyframes(path, kf)
+			err := getKeyframes(path, kf, sha)
 			if err == nil {
 				saveInfo(save_path, kf)
 			}
@@ -95,7 +91,7 @@ func GetKeyframes(sha string, path string) *Keyframe {
 	return ret
 }
 
-func getKeyframes(path string, kf *Keyframe) error {
+func getKeyframes(path string, kf *Keyframe, sha string) error {
 	defer printExecTime("ffprobe analysis for %s", path)()
 	// run ffprobe to return all IFrames, IFrames are points where we can split the video in segments.
 	// We ask ffprobe to return the time of each frame and it's flags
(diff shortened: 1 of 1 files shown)


transcoder/src/keyframes.go
```diff
@@ -164,10 +172,46 @@ func getKeyframes(path string, kf *Keyframe) error {
...
+	}
 	kf.add(ret)
 	if done == 0 {
 		kf.info.ready.Done()
 	}
 	kf.IsDone = true
 	return nil
 }
+
+func getFileDuration(path string) (float64, error) {
```
zoriya, line 193: You should use GetInfo from the `info.go` file. It has a cache and is already called before accessing keyframes.

transcoder/src/keyframes.go
```diff
@@ -164,10 +172,46 @@ func getKeyframes(path string, kf *Keyframe) error {
...
 		}
 	}
+	if len(ret) == 0 {
+		duration, err := getFileDuration(path)
+		if err != nil {
+			return err
+		}
+		ret = make([]float64, 0, int((duration/2)+1))
+		for segmentTime := float64(0); segmentTime < duration; segmentTime += 2 {
+			ret = append(ret, segmentTime)
```
zoriya, line 182: You don't need to append, you can use `ret[i] = i *2` if you set the len instead of the cap before.

This is more of a nit, you can keep like this.

transcoder/src/keyframes.go
```diff
@@ -135,6 +139,10 @@ func getKeyframes(path string, kf *Keyframe) error {
 		x := strings.Split(frame, ",")
 		pts, flags := x[0], x[1]
 
+		// true if there is no keyframes (e.g. in a file w/o video track)
+		if pts == "N/A" {
+			break
```
zoriya, line 144: Instead of breaking and making the normal path harder to read, I'd rather have a `return getDummyKeyframes(...)`

transcoder/src/keyframes.go
```diff
@@ -26,6 +27,9 @@ type KeyframeInfo struct {
 func (kf *Keyframe) Get(idx int32) float64 {
 	kf.info.mutex.RLock()
 	defer kf.info.mutex.RUnlock()
+	if len(kf.Keyframes) == 0 {
```
zoriya, line 30: Is this needed? The dummy keyframes should be created at this point, no?

zoriya (changes requested):

transcoder/src/keyframes.go
Arthi-chaud: oops, forgot to remove this (for test purposes)

transcoder/src/keyframes.go
Arthi-chaud: Wrote a dedicated `getDummyKeyframes` that would call `GetInfo`. However, I found out that audio-only files with an embedded image DO have a single keyframe (at time 0). So kept this `break` as it is, and completed the condition after the loop. Let me know what you think

transcoder/src/keyframes.go
zoriya: You should delete the now unused `getFileDuration`

transcoder/src/keyframes.go
zoriya: good catch

zoriya (approved):

Issue #357, opened 2024-03-25
Add helm charts


zoriya:
### Feature description

Steps for scalling (pseudo related)
 - [x] Move sql migrations code out of the API (#358)
 - [ ] See how to handle meilisearch schema
 - [x] Split scanner fs-monitor and item scan (use rabbitmq work queue to sync them) (#410)
 - [ ] #579
 - [x] Use a named queue for autosync (right now, I think that two instances of autosync would try to update the external service twice) (#435)
 - [ ] #580
 - [ ] #581
 - [x] Use traefik as a reverse proxy (#450)

bo0tzz:
For the [immich helm chart](https://github.com/immich-app/immich-charts), I used the [common-library](https://bjw-s.github.io/helm-charts/docs/common-library/) chart from bjw-s as a base, which helped get a chart together much quicker.

onedr0p:
> Rather than 1 container doing all the transcoding, you could create 1 pod per transcode stream.

That _may_ require a Kubernetes operator or for the application to at least be Kubernetes aware and have proper RBAC setup for the app to create transcoding pods.

bo0tzz:
I'm not particularly familiar with `HorizontalPodAutoscaler`s or KEDA, but maybe if one of those was involved scaling could be done without many changes on Kyoo's side?

zoriya:
I do want to support transcoder scaling and distributed transcodes. Not sure what is needed for that, I just finished reading the getting started of k8s x)

onedr0p:
Feel free to reach out if you need any help, I joined your discord 😄

atmosx:
> > Rather than 1 container doing all the transcoding, you could create 1 pod per transcode stream.
> 
> That _may_ require a Kubernetes operator or for the application to at least be Kubernetes aware and have proper Kubernetes RBAC setup for the app to create transcoding pods. It would be neat but might be outside the scope of this issue.

For HPA an operator is not required. You can setup auto-scaling based on CPU metrics, which come out of the box with any modern k8s distribution. An operator handles way more than autoscaling (e.g. automated DB backups/restore, version upgrades, storage expansion, maintenance tasks, etc.)

MegaShinySnivy:
I'd be willing to help test this chart and by extension Kyoo itself--I have enough users to put it through home/small event use (10ish users)

zoriya:
[...]

PR #560, opened 2024-07-11
add helm chart

acelinkio:
helm chart creation
closes: #357


acelinkio:
This is a work in progress, things that need to be added:
* Ingress
* Common Ingress annotations for URLRewrite. nginx, caddy, traefik, etc.
* Examples for GPU pass through
* Examples for SSO

Long term:
* Reduce reliance upon file storage for transcoder
* Reduce reliance upon file storage for back
* Image creation for matcher

chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,107 @@
+{{/*
+Create kyoo autosync name
+*/}}
+{{- define "kyoo.autosync.fullname" -}}
+{{- printf "%s-%s" (include "kyoo.fullname" .) .Values.autosync.name | trunc 52 | trimSuffix "-" -}}
```
zoriya, line 5: Why 52 chars?

chart/templates/back/deployment.yaml
```diff
@@ -0,0 +1,171 @@
...
+          {{- with .Values.back.kyoo_back.resources }}
+          resources:
+            {{- toYaml .Values.back.kyoo_back.resources | nindent 12 }}
+          {{- end }}
+          {{- with .Values.back.kyoo_back.containerSecurityContext }}
+          securityContext:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            {{- with .Values.media.volumeMounts }}
```
zoriya, line 150: The back does not need the media volume

chart/templates/matcher/deployment.yaml
```diff
@@ -0,0 +1,115 @@
...
+            - name: KYOO_APIKEYS
+              valueFrom:
+                secretKeyRef:
+                  key: {{ .Values.kyoo.apikey.apikeyKey }}
+                  name: {{ .Values.kyoo.apikey.existingSecret }}
+            - name: KYOO_URL
+              value: http://{{ include "kyoo.back.fullname" . }}:5000
+            - name: LIBRARY_LANGUAGES
+              value: {{ .Values.kyoo.languages }}
+            - name: THEMOVIEDB_APIKEY
```
zoriya, line 59: It's missing the `TVDB_APIKEY` & `TVDB_PIN` env vars

chart/templates/scanner/deployment.yaml
```diff
@@ -0,0 +1,123 @@
...
+            - name: KYOO_APIKEYS
+              valueFrom:
+                secretKeyRef:
+                  key: {{ .Values.kyoo.apikey.apikeyKey }}
+                  name: {{ .Values.kyoo.apikey.existingSecret }}
+            - name: KYOO_URL
+              value: http://{{ include "kyoo.back.fullname" . }}:5000
+            - name: LIBRARY_LANGUAGES
+              value: {{ .Values.kyoo.languages }}
+            - name: THEMOVIEDB_APIKEY
```
zoriya, line 61: The sanner does not need DB's api keys. Only the matcher uses them

chart/templates/transcoder/deployment.yaml
```diff
@@ -0,0 +1,111 @@
...
+            - name: GOCODER_PRESET
+              value: fast
+            - name: GOCODER_CACHE_ROOT
+              value: /cache
+            - name: GOCODER_METADATA_ROOT
+              value: /metadata
+            - name: GOCODER_PREFIX
+              value: /video
+            - name: GOCODER_SAFE_PATH
+              value: /data
```
zoriya, line 61: I have not seen volumes path hard-coded before, is it an oversight or by design?

zoriya: (talking about the `/cache`, `/metadata` & `/data`)

chart/templates/back/deployment.yaml
```diff
@@ -0,0 +1,171 @@
...
+          args:
+            {{- with .Values.back.kyoo_back.extraArgs }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+          env:
+            - name: TRANSCODER_URL
+              value: http://{{ include "kyoo.transcoder.fullname" . }}:7666
+            - name: PUBLIC_URL
+              value: {{ .Values.kyoo.address }}
+            - name: UNLOGGED_PERMISSIONS
```
zoriya, line 82: You are missing the `REQUIRE_ACCOUNT_VERIFICATION` & the `DEFAULT_PERMISSIONS` env vars.

chart/templates/scanner/deployment.yaml
```diff
@@ -0,0 +1,123 @@
...
+      containers:
+        - name: main
+          image: {{ .Values.scanner.kyoo_scanner.image.repository | default (printf "%s/kyoo_scanner" .Values.global.image.repositoryBase) }}:{{ default (include "kyoo.defaultTag" .) .Values.scanner.kyoo_scanner.image.tag }}
+          imagePullPolicy: {{ default .Values.global.image.imagePullPolicy }}
+          args:
+            {{- with .Values.scanner.kyoo_scanner.extraArgs }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+          env:
+            - name: SCANNER_LIBRARY_ROOT
```
zoriya, line 50: You are missing the `LIBRARY_IGNORE_PATTERN` env var

chart/values.yaml
```diff
@@ -0,0 +1,359 @@
...
+    extraVolumeMounts: []
+    # workaround until dedicated image is created
+    extraArgs: 
+      - matcher
+    extraEnv: []
+    image:
+      # workaround until dedicated image is created
+      repository: ghcr.io/zoriya/kyoo_scanner
+      tag: ~
+  replicaCount: 1
```
zoriya, line 222: Just to note, ATM the matcher's cache is not distributed (only an in-memory cache). I plan to add valkey/redis to fix this.

For now, adding replicas will make more requests to content db.

chart/values.yaml
```diff
@@ -0,0 +1,359 @@
...
+    readinessProbe: {}
+    resources: {}
+    containerSecurityContext: {}
+    extraVolumeMounts: []
+    extraArgs: []
+    extraEnv: []
+    image:
+      repository: ~
+      tag: ~
+  replicaCount: 1
```
zoriya, line 253: The scanner should never be replicated. It monitors the FS, so if two are running, it will push to rabbit events twice (like `file deleted at /vidoe/...` or `file created`).

chart/values.yaml
```diff
@@ -0,0 +1,359 @@
...
+        name: transcoder-storage
+      - mountPath: /cache
+        name: cache
+  volumes:
+    - name: transcoder-storage
+      persistentVolumeClaim:
+        claimName: transcoder-storage
+    - name: cache
+      emptyDir: {}
+  replicaCount: 1
```
zoriya, line 295: Distributed transcoder still is not done yet, I see two ways of making this work:
- Share the transcoder's info cache (but I think this would come with a performance cost to sync instances)
- Making sure the same video file is always handled by the same transcoder + sending messages to others transcoder when a client starts a stream (so others transcoder can know the user left & cleanup the leftover transcode process).

For now (and as long as we don't do the first idea), the /cache should NOT be shared between instances.

chart/templates/transcoder/deployment.yaml
zoriya: Seems like it's by design after reading the `values.yaml`.

chart/templates/matcher/deployment.yaml
```diff
@@ -0,0 +1,115 @@
...
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "kyoo.matcher.serviceAccountName" . }}
+      containers:
+        - name: main
+          image: {{ .Values.matcher.kyoo_matcher.image.repository | default (printf "%s/kyoo_matcher" .Values.global.image.repositoryBase) }}:{{ default (include "kyoo.defaultTag" .) .Values.matcher.kyoo_matcher.image.tag }}
+          imagePullPolicy: {{ default .Values.global.image.imagePullPolicy }}
+          args:
+            {{- with .Values.matcher.kyoo_matcher.extraArgs }}
```
zoriya, line 46: The matcher should contain the `macher` arg, without it it starts the scanner.

zoriya (commented):
Thank you for taking the time to do this! It's a bit overwhelming for me, so my review is probably not the best you'll get.

I think an ingress would be needed for it to work, no? Right now I think SSR should work but the android app & navigating on the front should fail since it will not be able to reach the back.


About the matcher image, do you think it's necessary? Most of the code is common between the scanner/matcher so it felt counterintuitive to make 2 images. I think loki uses the same pattern of having two binaries in the same image and using the `command` or `arg` flag to launch the one you want.

Ideas for later:
- The front assumes the api is available under `${front-url}/api`, maybe this assumption should be removed and allow users to configure how the front should reach the back? Might not be that configurable due to CORS & might make it harder for the mobile app to connect to kyoo (right now the mobile app takes the frontend url and adds /api to connect).

chart/templates/matcher/deployment.yaml
acelinkio: The templating here allows us to rely upon sane defaults inside of the `values.yaml` file.  In that file, the value is set to

```yaml
matcher:
  kyoo_matcher:
    extraArgs: 
      - matcher
```

if/when we create a dedicated matcher image, that value can be unset/overwritten.

chart/templates/scanner/deployment.yaml
acelinkio: I tried not to add every single configuration on the first pass.  The power user can set individual environmental variables via `scanner.kyoo_scanner.extraEnv`.  

Alternatively we can expose general settings via kyoo.languages or kyoo.unloggedPermissions.  Definitely would need some help picking out the most important ones to showcase.

chart/templates/back/deployment.yaml
acelinkio: similar to the scanner comment.

I tried not to add every single configuration on the first pass. The power user can set individual environmental variables via back.kyoo_back.extraEnv.

Alternatively we can expose general settings via kyoo.languages or kyoo.unloggedPermissions. Definitely would need some help picking out the most important ones to showcase.

chart/templates/transcoder/deployment.yaml
acelinkio: Trying to have everything be dynamic is challenge.  This felt like a safe place to hardcode and set expectations about where mounts are.

chart/templates/_helpers.tpl
acelinkio: Updated to 63.

chart/values.yaml
acelinkio: Added a comment stating it does not currently support multiple replicas.

chart/values.yaml
acelinkio: With this setup, the cache is stored between instances.  Need to do some more reading on best way to ensure each replica has its own cache... 

could just remove the mount and have it use the disk overlay.

chart/templates/scanner/deployment.yaml
acelinkio: Removed setting

chart/templates/back/deployment.yaml
acelinkio: Will remove tomorrow and also update the DIAGRAMS.md (probably in a separate PR)

acelinkio:
Will definitely add an ingress.  I have moved onto using GatewayAPI and sunset ingresses from my cluster.  Need a little bit of time to write the template and also want to find the annotations for the URLRewrite that is needed.  GatewayAPI's HTTPRoute handles that in spec.  


```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: kyoo
  namespace: kyoo
spec:
  parentRefs:
  - name: internal
    namespace: gateway
    sectionName: https
  hostnames:
  - "kyoo.mydomain.com"
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /api
    backendRefs:
    - name: kyoo-back
      port: 5000
    filters:
      - type: URLRewrite
        urlRewrite:
          path:
            type: ReplacePrefixMatch
            replacePrefixMatch: /
  - matches:
    - path:
        type: PathPrefix
        value: /
    backendRefs:
    - name: kyoo-front
      port: 8901
```

I am 50/50 on matcher.  Right now its functional and that's all that matters.

Ideally there is as few hops between the client and server as possible.  I know some folks bypass their nginx/traefik/etc when hosting Plex.  That's also the reason why `KYOO_URL` env is set to use http://kyoo-back:5000 for front/matcher/scanner.  No reason to go out then come back through an ingress or gateway.  Just go directly to through Kubernetes service.

bo0tzz:
Great stuff! For now I don't have any review comments, just the suggestion to yoink https://github.com/immich-app/immich-charts/blob/main/.github/workflows/pull_request.yaml - I've found it incredibly helpful to have when maintaining the Immich chart.

chart/templates/transcoder/deployment.yaml
zoriya: Kay, fine with me!

chart/templates/scanner/deployment.yaml
zoriya: I think it would be best to expose everything, I try to keep configuration options to a minimum.

It can be a challenge to find which service will read the option, making it harder to set up.

chart/values.yaml
zoriya: There is no plan to one day support replica for the scanner, it does not make sense to do so. I think removing the option all together would be better (with a comment explaining why it can't be replicated).

chart/templates/back/deployment.yaml
acelinkio: Removed backend mount and also updated diagram

chart/values.yaml
acelinkio: Resolving thread.  emptyDir creates a local cache that is not shared across pods.  This means if multiple replicas were running, each would have their own independent cache.  

Current cache writes to disk. Unsure what the performance requirements are for using a distributed cache.

chart/templates/scanner/deployment.yaml
acelinkio: Still mapping out what user inputs should look like.

Would prefer to have `SCANNER_LIBRARY_ROOT` be calculated from media.volumeMounts[0].mountPath[0] value.  For `LIBRARY_IGNORE_PATTERN` I would likely create variable inside of `media.ignorePattern`

acelinkio:
Hey @zoriya  In top level readme, I see there is a bullet point for Cloud Native that points at the issue this PR targets.  The helm chart itself is a large item and will have some additional follow up items including adding tests to CI/publishing helm charts to a helm registry and OCI registry.  Some of the transcoder logic could potentially be handled with infrastructure, but not depend on solving to create a helm chart.  Can the distributed processing be the focus of another issue? 

This weekend I will address the remaining comments and add an ingress.

zoriya:
Yes, of course, I should probably make each item its own issue & create a meta issue

chart/templates/matcher/deployment.yaml
acelinkio: added support.  resolving

chart/templates/back/deployment.yaml
acelinkio: Added those settings. closing

chart/templates/scanner/deployment.yaml
acelinkio: Added setting for configuring `LIBRARY_IGNORE_PATTERN` . closing

acelinkio:
Update the helm chart to address all of the comments.  Chart is ready for review.

There is a some room to document Kubernetes specific settings such as GPU passthrough.  I do not have the hardware right now to verify any examples I create work.

chart/values.yaml
```diff
@@ -319,15 +327,31 @@ meilisearch:
...
+        # https://github.com/zoriya/Kyoo/issues/536
+        kyoo_back.sql: |
+          CREATE DATABASE {{ .Values.global.postgres.kyoo_back.database }} WITH OWNER {{ .Values.global.postgres.infra.user }}; 
+          \connect {{ .Values.global.postgres.kyoo_back.database }};
+          CREATE SCHEMA IF NOT EXISTS data AUTHORIZATION {{ .Values.global.postgres.infra.user }};
+        kyoo_transcoder.sql: |
+          CREATE DATABASE {{ .Values.global.postgres.kyoo_transcoder.database }} WITH OWNER {{ .Values.global.postgres.infra.user }};
+          \connect {{ .Values.global.postgres.kyoo_transcoder.database }};
+          REVOKE ALL ON SCHEMA public FROM PUBLIC;
+          CREATE SCHEMA IF NOT EXISTS data AUTHORIZATION {{ .Values.global.postgres.infra.user }};
```
zoriya, line 375: does thsi means you expect the schema's name should be the same as the user?

i guess i should add a config option for that?

chart/values.yaml
```diff
@@ -97,10 +97,17 @@ media:
...
+  tmdb:
+    enabled: true
+    apikeyKey: tmdb_apikey
+    existingSecret: bigsecret
+  # TVDatabase
+  tvdb:
+    enabled: true
+    apikeyKey: tvdb_apikey
+    pinKey: tvdb_pin
+    existingSecret: bigsecret
```
zoriya, line 116: i think you would need two exitstingSecrets (one for api key & one for pin key, right?)

zoriya:
I reviewed, but I'm mostly just another pair of eyes as I'm not too familiar w/ k8s yet. I don't have a k8s env to test this so I'm trusting your tests.

tbh hwaccel enabled k8s can wait a future PR, this is already a huge achievement! Thanks for tackling this!

It would be great if you could update the main readme.md & installation.md to talk about this chart. Adding a note about no gpu passthrought yet on the chart/readme would be a good idea also

chart/values.yaml
acelinkio: This just creates a schema named `data` and assigns the user `kyoo_all` as the owner.   Your application still creates a gocoder schema--but i did confirm it starts with this approach. 

Following best practices for database roles is easier done with a Kubernetes operator.  Zalando for instance will create database roles for OWNER, READWRITE, and READONLY.  Then assign users into those roles. If using multiple schemas, best practice is to also create OWNER, READWRITE, and READONLY per schema that can be assigned to users.  Having a database per microservice is just easier and avoids unneeded creation of additional roles.

In this implementation, we are creating one postgres instance using bitnami's helm chart.  The instance on setup will create two database based upon  `.Values.global.postgres.kyoo_back.database` and `.Values.global.postgres.kyoo_transcoder.database`.  kyoo_back requires the use of public schema as mentioned in ttps://github.com/zoriya/Kyoo/issues/536 so we need to leverage it.  kyoo_transcoder creates a schema named data to avoid ever using public then facing permission issues.

I get why you are having the application create the schema, because it's easier to work with in docker-compose.  My opinion is that those tasks should be part of the infrastructure setup. Application should just assume they already exist--this avoids the application needing to be aware of roles configuration and permission issues in general.

chart/values.yaml
acelinkio: They are both expected to come from the same existingSecret.  All references at same level/place of the map use the same existing secret.  That secret just has multiple keys.  Chart readme includes an example that just uses a single secret for specifying all the key/value pairs.

Resolving this conversation.  Please reopen if you have any additional questions on this particular part.

acelinkio:
GPU Pass through support is here in the chart.  I just don't have an example of how to exactly configure that pass through.

Best practice is to consume helm charts that are published to an artifact repository.  I will be out for a couple weeks and wont have time to add workflows for lint/testing/publishing to a chart repository.  Once charts are published to an artifact repository he documentation and installation instructions should be updated.

chart/values.yaml
zoriya: the application creates them if they don't already exist, i don't think it conflicts with infra management of schemas, does it?

acelinkio: I see where you’re coming from, and it makes sense that having the application create schemas as a fallback can provide extra resilience. However, my main concern is that it can blur the lines of responsibility between the infrastructure and the application. When the application manages schema creation, it introduces potential complexities with role management and permissions that might be better handled during setup. 

If the infrastructure is consistently responsible for creating the schemas, then application deployments become cleaner since they assume everything is pre-configured. But I agree that having the application create schemas as a safety net can still work as long as it doesn’t lead to conflicts or inconsistencies down the road.

zoriya: I don't think we have another good option to keep a simple docker-compose deployment. Do we?

acelinkio: docker-compose based can be kept inline with this helm chart based initialization.  https://github.com/docker-library/docs/blob/master/postgres/README.md#initialization-scripts

have the postgres instance created without a default database, then use the following script to generate the databases.

```sql
CREATE DATABASE kyoo_backend WITH OWNER kyoo_all; 
\connect kyoo_backend;
CREATE SCHEMA IF NOT EXISTS data AUTHORIZATION kyoo_all;
CREATE DATABASE kyoo_transcoder WITH OWNER kyoo_all;
\connect kyoo_transcoder;
REVOKE ALL ON SCHEMA public FROM PUBLIC;
CREATE SCHEMA IF NOT EXISTS data AUTHORIZATION kyoo_all;
ALTER ROLE kyoo_all 
IN DATABASE kyoo_backend SET search_path TO "$user", public;
ALTER ROLE kyoo_all 
IN DATABASE kyoo_transcoder SET search_path TO "$user", data;
```

Using this, you can avoid creating a schema in the application and just rely upon what ever is in the search_path.

zoriya: The initialization scripts are only run once when the volume is created. This means it's impossible to add another schema when we add a new service. It doesn't seem like a good solution. I guess we could add an env var to disable schema handling in the app and let the infrastructure handle it tho.

acelinkio: I hear you that it is a pain mamaging resources.  but I also think it is unreasonable for Kyoo to start specializing in configuring postgres.

A way to address this is to use advanced tooling like postgres operator which will handle the lifecycle of the databases and also roles/permissions.  

example snippet for configuring a database named mealie, a schema named data, and will also create database level user/roles for owner/read-write/read-only.

source: https://github.com/acelinkio/argocd-homelab/blob/main/manifest/postgres.yaml#L28

```yaml
    mealie:
      defaultUsers: true
      extensions: {}
      schemas:
        data:
          defaultRoles: false
          defaultUsers: false
```

As always, when using external subchart configuration in charts is best effort.  For example, when I updated the Meilisearch version from 0.7.0 to 0.8.0, that was a breaking change that required manual intervention.

zoriya: I understand the want to manually control postgres, but most docker-compose users don't care (or know) how to do this. This is unreasonable to ask them to manually create databases and schema. The default behavior for docker-compose needs to work without any manual efforts, and future updates needs to just work. This is also one of the reasons the default docker-compose is keeping postgres at version 15.

I'm open to any modifications that could make infra control easier for kube (or for the DX) but docker-compose users should not have to know or do anything.

bo0tzz: Honestly - I understand why you're going with multiple databases, but if simplicity like that is a goal, you're shooting yourself in the foot by taking that approach.

acelinkio: This is typically what happens with microservice architecture. There is trade offs to each approach.  One way to avoid initialization complexity is to do a postgres instance per microservice.  The skilled power user can do consolidation if they wish.

joryirving (approved):
I think this is a solid attempt. I haven't had time to test it out myself, but it looks good.

chart/values.yaml
zoriya: I feel like a default schema by application is a sane default, if someone wants to customise this they could override the schema via env var (or disable custom schema handling from the app)

something like

`POSTGRES_SCHEMA` that would default to `gocoder` but you could use anything.
If you specify `POSTGRES_SCHEMA=disabled` it would just use the default `search_path`.

What do you think?

zoriya: I merged the `POSTGRES_SCHEMA` handling!

jess-sol:
I decided to give this helm chart a shot for installing Kyoo, and have initially run into thees small issues:
- Is `.Values.media.volumeMounts` missing from kyoo backend's deployment?
- Ingress doesn't work without removing the /api prefix with a rewrite annotation (which only nginx ingresses support). Maybe it's worth adding a way to configure `UseBasePath` with an environment variable to Kyoo to deal with this? (This way, when ingress is enabled, a base path of /api can be specified for the backend)
- media pvc in readme example should be `ReadWriteMany`?

I'm still sorting through bugs, and making changes I think improve the chart. But I'm also still coming up to speed on Kyoo in general, so wanted to make sure these issues aren't just misunderstandings on my part.

zoriya:
The backend doesn't require medias, only the scanner & the transcoder do.
Traefik also supports removing the /api prefix, but this should indeed be configurable. I plan to correct that with the v5.
Kyoo never writes inside the media pvc, it can be ReadOnly.

Thanks for testing the chart & kyoo! Feedback is very valuable!

acelinkio:
I had a small typo that was fixed in [`37fe57f` (#560)](https://github.com/zoriya/Kyoo/pull/560/commits/37fe57fdf40cb8353b297290b4b5944f45b0489d) and caused an issue submitting the kyoo-backend deployment.

> I decided to give this helm chart a shot for installing Kyoo, and have initially run into thees small issues:
> 
>     * Is `.Values.media.volumeMounts` missing from kyoo backend's deployment?
> 
>     * Ingress doesn't work without removing the /api prefix with a rewrite annotation (which only nginx ingresses support). Maybe it's worth adding a way to configure `UseBasePath` with an environment variable to Kyoo to deal with this? (This way, when ingress is enabled, a base path of /api can be specified for the backend)
> 
>     * media pvc in readme example should be `ReadWriteMany`?
> 
> 
> I'm still sorting through bugs, and making changes I think improve the chart. But I'm also still coming up to speed on Kyoo in general, so wanted to make sure these issues aren't just misunderstandings on my part.

I'm not sure where ReadWriteMany is setup.  I don't see it anywhere in the code base.  Maybe you are looking at an old commit?

The working example is in the readme in the charts directory.  https://github.com/acelinkio/Kyoo/tree/feature/helmchart/chart#quickstart.  The chart does not create PVCs, instead it expects to consume existing ones--mostly because there is no sane default that is friendly for everyone.  (Also note that transcoder and back storage are eventually moving away from leveraging the filesystem)

As for ingress rules... yes the application does rely upon a rewrite to work properly.  I have moved onto using Gateway APIs for managing rewrites, so I do not have an convenient way to test ingress rewrite annotations.  If you come up with some working annotations, will gladly those as defaults.

joryirving:
IIRC the media storage usually is bulk storage like an NFS mount, rather than a PVC to be consumed, unless you plan on storing your media library in cluster on a big PVC.

chart/values.yaml
acelinkio: Updated helm chart to support specifying schema for transcoder.  Also added docs that encourage users to not rely upon sub-charts.  

Resolving conversation.  Please reopen if there are any questions or concerns.

jess-sol:
@acelinkio 
For the media PVC I'm referring to the media PVC in the extraObjects array in the example in the readme. Its accessMode is set to ReadWriteOnce ([see here](https://github.com/acelinkio/Kyoo/blob/feature/helmchart/chart/README.md?plain=1#L73)), but it's mounted across multiple containers, so I thought it should be ReadWriteMany. For example, in my cluster I use CephFS to provide ReadWriteMany PVCs, one of which hosts my media.

As for the ingress, we need to add an option to the backend app which allows for specifying a subpath to serve on. This way we can avoid having to do a rewrite, which is ingress controller specific, and only supported by a subset of them. Maybe @zoriya can help get that in quickly. I'm not familiar with C# web applications, but it seemed like we need a way to configure `UseBasePath` from an environment variable.

That makes sense with backend not requiring the media volume, it should probably be removed from the [list of volumes for the deployment](https://github.com/acelinkio/Kyoo/blob/feature/helmchart/chart/templates/back/deployment.yaml#L164).

@joryirving Yes media is usually a file based storage system (like NFS or CephFS), but I've always seen it mounted using a CSI (such as the [NFS CSI](https://github.com/kubernetes-csi/csi-driver-nfs) or [cephfs csi](https://github.com/ceph/ceph-csi)). Have you seen it done another way?

joryirving:
Yes, I mount it directly to the container as an NFS mount.
```    
	persistence:
      media:
        type: nfs
        server: voyager.internal
        path: /mnt/data
        globalMounts:
          - path: /data
            subPath: media
            readOnly: true
```
https://github.com/joryirving/home-ops/blob/589dc78908d506590c08e45df27322f1b0b457c6/kubernetes/main/apps/media/kyoo/app/helmrelease.yaml#L305

acelinkio:
Hey @jess-sol ,

I can update the example.  For testing purposes, I just kubectl cp'd files into the mount and required write access.  All kyoo applications that mount media only require read access as mentioned above.  I went ahead adjusted to `ReadOnlyMany` emphasize this.

Thanks for catching the media volume in the back.  Removed the volumemount earlier, but missed the volume.  Fixed.

The rewrite is not a big deal.  Getting everything right on ingress controllers is a pain, but thankfully gateway api's make it a lot simpler going forward.  Still not entirely sure how to template an httproute.

acelinkio:
Helm chart is ready to go.  

In another MR we can introduce CI workflows.  One of the things we'll need to do is add a workflow to publish helm charts to a artifact repository.  That's when we can discuss releases/versioning (does the helm chart directly follow Kyoo's versioning or independently versioning).  

Once helm charts are published, update docs with detailed installation instructions.

zoriya:
Thanks so much @acelinkio for your work on this! Do you want to keep it a bit here for some testing, or should I merge this now?

jess-sol:
I think it'd be worth it to take one more pass on the values.yaml file and
simplify it. I have some changes I'd like to try propose before people
start relying on it.

acelinkio:
Hey.

I am comfortable merging and having follow up MRs to make adjustments.  Helm chart is not being published anywhere so its still very mutable.  

@jess-sol, definitely down for feedback.  Could you outline some of the suggestions?  From a deployment standpoint the chart is pretty minimal touch needed.  There are just a lot of config options exposed in values.yaml

jess-sol:
@acelinkio A few examples I was considering:
1. Rework secret references in values.yaml to be actual [SecretReferences](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#secretreference-v1-core), [like this](https://github.com/acelinkio/Kyoo/commit/0fbe33734c3f262f10caaa3eb9d4341e5d3b2236).
2. Quote environment variable values, I actually ran into a bug where an environment variable (with an interpolated value) wasn't quoted, and broke rendering the template.
3. Better comments on what each value does. For example, be more clear why `adminPasswordKey` in the bitnami Postgres chart can't be a variable, and must be set explicitly (it's because of a quirk in upgrade behavior in the Bitnami chart).
4. Refactor `global.postgres`, I don't think exposing the names of schemas as subkeys of the databases makes sense (for kyoo_migration and kyoo_back). I'd rather expose them in a more generic way, maybe `migrationsUserSecretRef`, and `appUserSecretRef`? Then the nested kyoo_transcoder could be not nested as well.
5. A bit of a nit, but I don't think variables should be prefixed with `kyoo_`. I get it's to match the database/schema naming, but I think if the `global` variables are more abstract it'd simplify usage of the chart, especially when using with an external resources (such as DB, meilisearch, rabbitmq, etc.). I'd like to test installation with all external resources too, to make sure it's as simple as possible (and possibly write docs for what DBs/schemas need to exist).
6. Ingress is still broken, and requires either custom annotations, or a change to the backend's source to support a new configuration.

I don't have a ton of time right now, but I was going to put these changes together and submit a PR into your PR. I wanted to wait to merge because if any of the changes I propose around values.yaml variables changes get merged in, it'll break any existing installations of the Helm chart

acelinkio:
1. Adopting SecretReferences adds complexity/verbosity to the values.yaml.  That change would add another ~40 lines without really adding much.  The assumption that related information like user/password live in the same secret is sane.  Also the subcharts do not support that approach.

2. That setting in https://github.com/zoriya/Kyoo/pull/560/commits/37fe57fdf40cb8353b297290b4b5944f45b0489d.  However I would definitely be open to a PR to make the quotes more uniform.  I think in some places its `{{ $variable| quote }}" and others its quotes around the templating.

3.  I am definitely down for adding additional comments into the values.yaml.  A style choice was made around credentials, everything should follow Kubernetes best practices of using secrets instead of plaintext.  That is the reason why `adminPasswordKey` leverages a secret, because that's how Kyoo also leverages them.

4.  Postgres Databases, Schema Management, and Extensions should all managed by who/what is maintaining the Postgres Cluster/Instance.  Note that those are two distinctly separate databases. `global.postgres.kyoo_back.database` and `global.postgres.kyoo_transcoder.database`.   

kyoo_transocder has a schema configuration `global.postgres.kyoo_transcoder.schema` because application defaults to creating its own schema and that is not desirable.  Setting that setting to disabled, means the application relies upon the search_path.  TLDR it is a compromise between ease of use and not creating a postgres instance per database.   The context around that setting is in this https://github.com/zoriya/Kyoo/pull/560#discussion_r1723352738.

5.`kyoo_autosync`, `kyoo_back`, `kyoo_front`, `kyoo_matcher`, `kyoo_migrations`, `kyoo_scanner`, `kyoo_transocder` are all names of the actual image artifacts being deployed.  https://github.com/zoriya/Kyoo/blob/master/DIAGRAMS.md might help make things be clear.

The settings that you dislike are in global postgres/meilisearch/rabbitmq.  The references to `kyoo_` in those places is to refer to the actual image/process that is consuming those variables.   Postgres is slightly more complicated because there is a database named `kyoo_back` and that database has two workloads connecting `kyoo_migrations` and `kyoo_back`.  There is also a separate database named `kyoo_transcoder` which has one workload consuming it `kyoo_transocder`.

If there was one thing I could change, would be renaming`kyoo_migrations` to `kyoo_back_migrations` to make it clear its only managing one specific database.

6. Ingresses are disabled by default.  To get ingresses working a rewrite annotation.  That annotation is specific for each ingress controller implementation.  We can definitely add in several annotations for common ingress controllers such as nginx-ingress, caddy, traefik, etc--but I do not have any of those deployed or changed.   I do not have any of those currently installed and when I looked at nginx-ingress--I could only find the 'unsecure' way of doing it.


PR for point number 2, 3, and 6 would be appreciated but those would not incorporate breaking changes.

zoriya:
I'll merge this for now, i prefer the master to be the working document than starting to have PR to PRs.

It could be nice to have a meta issue with work needed before we publish the helm chart somewhere.

Thank you for your investment @acelinkio & @jess-sol! Thanks @joryirving & @bo0tzz for your reviews!

PR #561, opened 2024-07-11
update diagrams based upon helm updates

acelinkio:
Change removes media/back relationship.  Media files do not need to be mounted to kyoo_back

Reorganized the graphs to render a little nicer and be less noisy.


acelinkio:
Converting to draft because my local render of the diagrams looks different from Github's renderer

zoriya:
Thanks for the corrections!

PR #562, opened 2024-07-12
chore(renovate): remove biome config

onedr0p:
As discussed in https://github.com/zoriya/Kyoo/pull/552


zoriya:
Thanks, it was on my todo list!

PR #573, opened 2024-07-25
Add auth spec

### LINKED ISSUE #346: OIDC using a public auth server
Right now, every instance of kyoo needs their own api keys for every oidc provider. This is nice and will never be deprecated but for users that don't want t

zoriya:
View the rendered spec here: https://github.com/zoriya/Kyoo/tree/feat/auth/auth

The Federated part of the doc would close #346

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -1,0 +1,141 @@
+# Auth
+
+## Features
+
+- Not an oauth provider/no login page (as in you don't redirect to this, you create your own auth page)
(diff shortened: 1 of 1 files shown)


zoriya:
To convert opaque tokens to jwt, the api gateway/reverse proxy should do it (see phantom token doc in readme.md). I don't think there is a built in way to do that with traefik so we might need to write a custom plugin? or maybe using something else

PR #582, opened 2024-08-04
Front: Fix Media Metadata

Arthi-chaud:
The `MediaSessionManager` seems to be broken:

Screenshot 2024-08-04 at 19 46 57

This PR provides a fix (there was a mismatch between props of `<Video/>` and `<MediaSessionManager/>`)

Additionally, I took the liberty to rework the format of the data passed to `<NativeVideo/>` and `<MediaSessionManager/>`:

- We pass the poster (of the movie or the parent show) instead of the thumbnail
- The pass a _subtitle_, only for episodes, which will be the name of the show
  - Fun fact, NativeVideo actually accept such a prop, didn't know that. For web, it will be considered as the album name

### Screenshots

Screenshot 2024-08-04 at 19 46 45
Screenshot 2024-08-04 at 19 47 32


zoriya (approved):

zoriya:
Thanks for yet another great contribution!

PR #585, opened 2024-08-07
Front: Home Page Header: Use Logo

Arthi-chaud:


zoriya:
closing due to staleness (the front was almost completly rewritten x))

Issue #581, opened 2024-08-04
Route users to the same transcoder instance if they are requesting an video already being handled by an instance


zoriya:

acelinkio:
Closing this issue as it is being discussed in https://github.com/zoriya/Kyoo/pull/914

PR #586, opened 2024-08-07
Use a database for the transcoder

zoriya:
Steps needed for #542 & multi-video support (in the same file).

A rudimentary multi-video m3u8/paths are made available, but they are not tested & no ui button was made for this PR.

Each part of the code is versioned (info, keyframes, sub extraction & thumbnails creation). Simply incrementing the version constant in the code will allow to update only this data.

Keyframes are now handled at the stream level instead of the file level. This is needed for fmp4 support (#542) & allows better resource management (FU PR).


zoriya:
This PR is a breaking change for downstream users of gocoder as the video path changed (from `/:path/:quality/{index/segment}` to `/:path/:video-idx/:quality/{index/segment}`). This path is only exposed through the master.m3u8, so this should be transparent, tho.

@Arthi-chaud for blee

zoriya:
This PR also lays the groundwork to support replications. A new `RunLock` struct was created and guards long-running tasks from running multiples times. This lock could be distributed via redis.

For the tracker (knowing what user watches & the time), this could also be distributed via redis.

The streams/segments should not be shared & a router rule could be used to maximize caches hit (aka making an instance responsible for a file as specified in #581)

PR #587, opened 2024-08-08
Add external subtitles support

### LINKED ISSUE #253: Parse external subtitles
### Feature description

The scanner should be able to retrieves external subtitles files and the /info of the trancoder should list them as well (maybe the scanner can ignore them and the transcoder could just check for them?)

zoriya:
Also wait for extraction to finish before returning attachments to prevent race conditions.

For now, CJK subtitles have issues since the webasm env does not have a compatible font. (it looks like the image below)

image

Closes #253


zoriya:
I'll merge this for now, we'll make subs in a FU pr (it at least works for .srt, .vtt & latin .ass)

PR #590, opened 2024-08-10
Front: Add Gradient behind navbar if transparent

### LINKED ISSUE #251: The navbar on the home page and the details page is illegible with some images
### Feature description

With light images, we can't see the search bar or the browse button

Arthi-chaud:
Adds a gradient behind navbar when transparent.

There are a lot of things that I am not happy with, but I hope this can serve as food for though for the future:

- Setting the tooltip's zIndex at 10 is arbitrary
  - Though it's needed so that tooltips do not appear behind the navbar (for example on the search bar)
- I am not a fan of how it looks. To me, this gradient makes the navbar look bulky
- I don't like the way we inject a 'background' into the navbar

# Screenshots

Screenshot 2024-08-10 at 15 30 03
Screenshot 2024-08-10 at 15 30 09

Closes #251

### DIFF
--- a/front/apps/web/src/pages/_app.tsx
+++ b/front/apps/web/src/pages/_app.tsx
@@ -201,7 +201,7 @@ const App = ({ Component, pageProps }: AppProps) => {
 												{...props}
 											/>
 										</ConnectionErrorVerifier>
-										<Tooltip id="tooltip" positionStrategy={"fixed"} />
+										<Tooltip id="tooltip" style={{ zIndex: 10 }} positionStrategy={"fixed"} />
 										<SetupChecker />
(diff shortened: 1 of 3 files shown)


zoriya:
Thanks for tackling this! It does look rather bulky, maybe setting a lower gradient (a gradient with transparent for example) would help.

For reference, here is the ImageBackground gradient: https://github.com/zoriya/Kyoo/blob/master/front/packages/primitives/src/image/index.tsx#L152-L154

Arthi-chaud:
There we go! It's much better: 
Screenshot 2024-08-10 at 19 38 47
Screenshot 2024-08-10 at 19 38 55

PR #592, opened 2024-08-11
Switch to -f hls instead of -f segment

zoriya:
Another step needed for #542

### DIFF
--- a/transcoder/src/keyframes.go
+++ b/transcoder/src/keyframes.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"log"
+	"math"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -12,7 +13,10 @@ import (
 	"github.com/lib/pq"
 )
 
-const KeyframeVersion = 1
+const KeyframeVersion = 2
+
+// In seconds, the spec recomands 6 but since we don't control keyframes we go over more often than not.
+const OptimalFragmentDuration = float64(5)
 
 type Keyframe struct {
 	Keyframes []float64
@@ -173,6 +177,7 @@ func getVideoKeyframes(path string, video_idx uint32, kf *Keyframe) error {
 
 	ret := make([]float64, 0, 1000)
 	limit := 100
+	last_frame := math.Inf(-1)
 	done := 0
 	// sometimes, videos can start at a timing greater than 0:00. We need to take that into account
 	// and only list keyframes that come after the start of the video (without that, our segments count
@@ -181,6 +186,9 @@ func getVideoKeyframes(path string, video_idx uint32, kf *Keyframe) error {
 	// We can't hardcode the first keyframe at 0 because the transcoder needs to reference durations of segments
 	// To handle this edge case, when we fetch the segment n0, no seeking is done but duration is computed from the
 	// first keyframe (instead of 0)
+	//
+	// since we switched to -f hls, duration is not computed anymore so we could hardcode the first timestamp at 0 but
+	// not changing it is easier.
 	for scanner.Scan() {
 		frame := scanner.Text()
 		if frame == "" {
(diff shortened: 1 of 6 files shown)


transcoder/src/keyframes.go
```diff
@@ -199,12 +212,18 @@ func getVideoKeyframes(path string, video_idx uint32, kf *Keyframe) error {
...
-		// to prevent segments of 0.2s but sometimes, the -f segment muxer discards
-		// the segment time and decide to cut at a random keyframe. Having every keyframe
-		// handled as a segment prevents that.
+		// The -f hls encoder will cut at the next keyframe found every X seconds.
+		// To be sure there is no mismatch between what -f hls outputs and what we think segments are, we need to make
+		// sure every keyframes is at least `OptimalFragmentDuration` away from the next.
+		//
+		// Note that this handling was not possible with the -f segment muxer, since if sometimes doesn't care about the
+		// -segment_times option and outputed segments of 0.2s. With -f segment, we needed to cut at every keyframe.
+		if fpts < (last_frame+OptimalFragmentDuration) {
```
zoriya, line 221: This might cause issues with the `-ss` handling of video items.

We might need to have two lists, one for segments and one for keyframes

zoriya: Specially if we have keyframes at 0,5,9.8,10. We might end a segment at 10 & start the next at 9.8

zoriya:
`-hls_time` doc said `Segment will be cut on the next key frame after this time has passed.` but that's a lie. It tries to keep an "average" segment length ([source](https://superuser.com/questions/1414310/ffmpeg-ignores-hls-time-argument)). For `-hls_time 5`, I got:
```
0-segment-0.ts - 5985
0-segment-1.ts - 3983
0-segment-2.ts - 5985
0-segment-3.ts - 3983
0-segment-4.ts - 5985
0-segment-5.ts - 3983
```

Not sure if this is deterministic, I don't know if switching to `-f hls` is possible...
There is no other ways of specifying when to split segments (looking at https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/hlsenc.c#L3147).

zoriya:
From https://x.com/JeWe37/status/1822774329152508398?s=19

it splits at the next keyframe after hls_time * [the number of segments written so far]
+/- an initial time that i dont know of

-hls_time seems like a no go

JeWe37:
Possible solution I came across while reading up on something related: MP4Box should probably be able to do the required muxing to HLS, with the [-bound](https://wiki.gpac.io/MP4Box/mp4box-dash-opts/#options) option and forcing keyframes as you already are. The idea is then that you can make the duration so small that MP4Box will split at every keyframe, because it's desperately trying to trying to fragment more finely than it actually can.

Hint from my own experience that took me a good while to figure out: Piping from ffmpeg to MP4Box is not entirely trivial. The only way I've found to do it is to use a named pipe, pipe the output from ffmpeg into that and read from it in MP4Box via `pipe://PATH`. Example:
```
mkfifo test.mp4
ffmpeg -to 3:00 -i /mnt/Anime/tv_shows/Oshi\ no\ Ko\ \(2023\)/Season\ 2/Oshi\ no\ Ko\ \(2023\)\ -\ S02E06\ -\ \[UNIVERSAL\].mp4 -force_key_frames "expr:gte(t,n_forced*2)" -sc_threshold 0 -s 640x480 -c:v libx264 -b:v 1536k -an -f mpegts - >> ../test.mp4
MP4Box -dash 6000 -segment-name 'seg_$Number$' pipe://../test.mp4 -out live.m3u8:dual:cmfc
```
The playlist only gets written at the end, but since you aren't using it and only need correct segments anyway that shouldn't be a concern.

Not entirely sure about the latency implications of this approach.

JeWe37:
Scratch MP4Box, this works with ffmpeg too? 
```
ffmpeg -i yourfile.mp4 -force_key_frames "expr:gte(t,n_forced*2)" -sc_threshold 0 -s 640x480 -c:v libx264 -b:v 1536k -an -hls_time 0.1 -hls_playlist_type vod -hls_segment_type fmp4 -hls_segment_filename "fileSequence%d.mp4" prog_index.m3u8
```

zoriya:
Splitting when we re-encode is not a problem since we control both keyframes & segments. The real issue is when we transcode (aka `-c:v copy`). We cannot specify keyframes manually and must relly on the existing ones.

I saw mp4box could be useful (especially w/ #542 in mind) but using another tool means having to decode (maybe even encode) each segment twice. This also means having to maintain two hwaccel handling.

JeWe37:
How can that be a problem then? So long as the muxed(pristine) case matches the transcoded cases all is well? Therefore so long as one can determine in advance how the muxed case will be split, one just needs to make sure the transcoded cases, where we do have full control, are split the same? What is the troublesome constraint?

As for MP4Box, yea, I was thinking fmp4 already, but I don't think it really presents much additional value compared to ffmpeg for the application here. I don't see how it would cause any double encoding/deciding, it can just handle the muxing alone, see that example I gave of how to pipe between it and ffmpeg.

zoriya:
The issue is that we don't know how the muxed stream is going to be split. If we find keyframes at 0,2,3,5,6 w

zoriya:
I thought about it again and it could work AS LONG AS we can reliably seek/start transcoding at a specific keyframes. Right now, we seek before the keyframes we are interested in and rely on the `-f segment` cutting it where we want.

`-ss` seek is reliable for transcode but in `-c:v copy` mode it can't only seek to keyframes and when we specify exactly a keyframe it sometimes pick the previous one (i think due to rounding). I think we could add a value to offset that but it was not the most reliable last time I checked. 

I'll try again later! Thanks for your help on this, I had given up!

PR #595, opened 2024-08-13
Front: Improve readability in menus

Arthi-chaud:
This is a draft, as I think there might be a better way to do this.

This improves the readability of menu items on hover.
(This is of web only btw)

### Before
Screenshot 2024-08-13 at 19 37 05
Screenshot 2024-08-13 at 19 38 01

### After
Screenshot 2024-08-13 at 19 36 40
Screenshot 2024-08-13 at 19 37 48


zoriya:
This is way better, thanks!

PR #602, opened 2024-08-21
Fix playback with avi files missing PTS

zoriya:
AVI doesn't really support H.264 ([source](https://stackoverflow.com/questions/54923062/why-do-i-lose-pts-information-when-encoding-to-h264)). AVI files sometimes don't contain PTS timestamps.

This means keyframe extraction succeed with no keyframes and starting a stream panics.

We can instead use DTS to get keyframe timing ([source](https://trac.ffmpeg.org/ticket/5184)).

### DIFF
--- a/transcoder/Dockerfile
+++ b/transcoder/Dockerfile
@@ -1,9 +1,9 @@
 # FROM golang:1.22 as build
-FROM debian:trixie-slim as build
+FROM debian:trixie-slim AS build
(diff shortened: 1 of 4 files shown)


zoriya:
genpts as an input flag for transcode source: https://trac.ffmpeg.org/ticket/6905

zoriya:
still need to test this w/ normal streams to ensure nothing is broken (my computer decided to not randomly give up hwaccel so not tonight)

PR #604, opened 2024-08-27
Fix audio segment miscount

zoriya:
Audio duration can be different from container's duration. This leads to miss count of segments and the player waits for it before ending the playback.

This read the audio duration from the packets instead.

note: ffprobe can't be used to get audio duration directly since some formats don't have this metadata (mkv for example).


zoriya:
This technically works but it's slow af. `read_interval` only seems to disable printing, as I don't see much time improvements when using it.

Issue #633, opened 2024-10-02
Helm Chart Finalization


acelinkio:
### Feature description

Follow up from #560 MR.  Here is a checklist of items to improve the helm chart further and prepare it for prime time!

- [x] Add CI workflows to test chart
- [x] Add CI workflows to publish chart to a Helm/OCI repository
- [x] ~~Add common ingress controller annotations for rewrite needed on ingress~~ (Will add docs if https://github.com/zoriya/Kyoo/issues/653 is not added before then)
- [ ] Add installation documentation (chart needs to be published)
- [x] Uniformly quote environment variables throughout the chart
- [x] Improve values.yaml comments/documentation

acelinkio:
Having some troubles with common ingress controller annotations for rewrite as there is no standard implementations across controllers.  

**nginx-ingress:** Most common ingress controller
* rewrite via `configuration-snippet` annotation is the commonly used method by many people.  However it is no longer enabled by default due to security concerns.  https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#configuration-snippet
* rewrite via `rewrite-target` is the alternative method.  However it requires that each path be split into its own ingress resource.  https://kubernetes.github.io/ingress-nginx/examples/rewrite/#rewrite-target

**haproxy:** 2nd most common
* rewrite via `path-rewrite` annotation.  The most flexible to implement because you just repeat the path rule.  https://www.haproxy.com/documentation/kubernetes-ingress/community/configuration-reference/ingress/#path-rewrite

**nginx inc:**
* rewrite via `rewrites` annotation.  Requires that you use the Kubernetes servicename in the annotation.  That is annoying because template functions can't be used inside of values.yaml  https://github.com/nginxinc/kubernetes-ingress/tree/main/examples/ingress-resources/rewrites

**traefik:**
* requires its own stripprefix custom resource.  https://doc.traefik.io/traefik/middlewares/http/stripprefix/#prefixes

**caddy:**
* does not appear to be a widespread enough ingress controller.  Docs lack examples for rewrite annotations.  https://github.com/caddyserver/ingress/blob/main/internal/caddy/ingress/annotations.go#L7-L8

The next api replacing ingresses, GatewayAPI, handles rewrites in a standardized way
```yaml
spec:
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /api
    backendRefs:
    - name: kyoo-back
      port: 5000
    filters:
[...]

PR #637, opened 2024-10-05
improve quote usage

acelinkio:
Relates to #633 

Makes quoting more uniform across the helm chart to avoid type issues.


zoriya (approved):

zoriya:
as always, thanks for your work on this!

PR #667, opened 2024-11-09
Feat: Add missing languages options

felipemarinho97:
Its 99% transalated but its not listed


felipemarinho97:
Unfortunately, I was not able to test the changes locally due to https://github.com/zoriya/Kyoo/issues/668

zoriya:
wait you're right, a lot of translations are not imported there! that's a huge oversight, we should find a way to include all translations without requiring to edit this file manually.

we could either:
- import all files in the directory
- use a remote fetch plugin (this would also allow translations to be updated without an update/release)

felipemarinho97:
@zoriya It doesn't seem like an easy task, for now I've added all languages ​​with translation > 30%. I was able to validate the component as well, without connecting to a backend.
Screenshot 2024-11-10 at 00 51 49

Issue #670, opened 2024-11-10
Cannot set prefered playback language for language variations (ex: pt-BR)


felipemarinho97:
### Kyoo's version

4.7

### What happened?

I am unable to set "Portuguese (Brazil)" as my prefered audio and subtitle language. The language is available in "langmap" but is hidden from the interface.

PR #671, opened 2024-11-10
Add language variations

felipemarinho97:
Closes https://github.com/zoriya/Kyoo/issues/670

### DIFF
--- a/front/packages/ui/src/settings/playback.tsx
+++ b/front/packages/ui/src/settings/playback.tsx
@@ -18,14 +18,14 @@
  * along with Kyoo. If not, see <https://www.gnu.org/licenses/>.
  */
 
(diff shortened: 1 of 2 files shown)


zoriya:
We might want to improve this by using smarter fallbacks (like falling back to pt if pt-bz is not available, or things like that).

felipemarinho97:
> We might want to improve this by using smarter fallbacks (like falling back to pt if pt-bz is not available, or things like that).

I totally agree!

A fallback would be great, but at best the preferred language should be a priority list. For example, if a user is bilingual, he or she could select two or more languages. And the service would try the 1st, 2nd, and so on. This definition is general enough to serve as a fallback between pt and pt-BR as well.

I believe that's how Netflix works.

Issue #597, opened 2024-08-14
v5 plans


zoriya:
I talked a bit about it on the discord or twitter, but it's time to put everything on text:

I'm planning a complete rewrite of the backend & database restructuration. This is needed to support #87 and #282. This will also make #463 (big thanks to @Arthi-chaud for the brainstorming there) or #549 possible. This is also a great occasion to tackle tech dept and bad decisions I made during the 5 years I've been working on this codebase.

I plan on writing diagrams and asking for feedback on the discord at most turning points of the code, so please give your feedback if you're interested!

To give a concrete vision, I plan to:
 - [x] Rewrite the auth and extract it to its own service
	 - [x] Write the spec #573
	 - [x] Write the service #610
	 - [ ] Handle #346 
 - [ ] Rewrite the API
 	- [x] Design core database logic (see below)
 	- [x] Design a new API to handle mixes of episode/movies/special/recaps/extra
 	- [x] Rewrite the existing API (elysia w/ bun - javascript)
 	- [ ] Add websockets as a core part of the API (elysia has great websockets support)
 	- [ ] Handle scaling (websockets needs custom scaling logic)
 - [ ] Update the scaning API
   - [x] Make the spec #678
   - [ ] Implement it
 - [ ] Probably merge `autosync` into the API

I'm going to explain each points and why I want to design Kyoo this way, as always feel free to give your opinion or ideas.

## Why a separate auth service

Kyoo has multiples HTTP services, for now we have the API & the transcoder. To ensure users have the correct level of permissions, all requests hit the API which does permission validation and then proxy the transcoder. This is bad for performances, scaling and DX. This idea with a centralized auth service is to have the reverse-proxy/ingress/gateway call the service and trade an opaque auth token for a jwt (see #573's phantom token part). The short-lived jwt will be used by downstream services (API, transcoder, scanner...) to check for permissions.

Having this service stand alone also makes it possible/simple to have #346.

The auth service could also be used by others applications (as long as they are compliant w/ the license).

## Why rewrite the API from scratch

The current backend is written in C# which lacks sum-types. Kyoo's logic often works on types like `Movie | Serie`, `Episode | Movie`, `Episode | Special` and so on. The lack of sum-types in C# makes it hard to work with, we have multiples interfaces and custom logic scattered everywhere to handle this well. 
[...]

Issue #87, opened 2022-03-22
Support library translations (name, descriptions, images of shows/movies...)


zoriya:
Titles and descriptions should be translatable and translations should be stored in the database.

bo0tzz:
https://weblate.org/ is a great tool for creating translations, and they offer a free cloud plan for open-source projects.

zoriya:
No, this issue is for fetching metadata in multiples languages from themoviedb and allow users to select which version they want. This includes things like titles, poster or descriptions of shows/movies...
The web/android app already supports translations, it uses simple json (stored in https://github.com/zoriya/Kyoo/tree/master/front/translations) for now tho. I could be nice to use a service for that

bo0tzz:
My bad, I misunderstood the feature description!

zoriya:
It was a pretty bad name, that's fair

zoriya:
Implemented in master, will be released with v5!

PR #677, opened 2024-11-14
Feat/localized images

felipemarinho97:
* Fix: Add .env setting to prevent Kyoo for always downloading images in the original language. (bug?)
* Feat: Add an easy-to-understand algotithim to pick the best available image from TMDB.
* Fix: Add `include_image_language` querystring on tmdb request to include more image options and prevent empty posters for less known/rare media or for when `{ISO-639-2}-{ISO-3166-1-alpha-2}` langcodes are used on setup (eg. pt-BR).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,7 +13,7 @@ LIBRARY_LANGUAGES=en
 # If this is true, kyoo will prefer to download the media in the original language of the item.
 MEDIA_PREFER_ORIGINAL_LANGUAGE=false
 # A pattern (regex) to ignore files.
-LIBRARY_IGNORE_PATTERN=".*/[dD]ownloads?/.*|.*\.(mp3|srt|jpg|jpeg|png|gif|bmp|tiff|svg)$|.*[Tt][Rr][Aa][Ii][Ll][Ee][Rr].*"
+LIBRARY_IGNORE_PATTERN=".*/[dD]ownloads?/.*|.*[Tt][Rr][Aa][Ii][Ll][Ee][Rr].*"
 
 # If this is true, new accounts wont have any permissions before you approve them in your admin dashboard.
 REQUIRE_ACCOUNT_VERIFICATION=true
--- a/.env.example
+++ b/.env.example
@@ -12,8 +12,8 @@ CACHE_ROOT=/tmp/kyoo_cache
 LIBRARY_LANGUAGES=en
 # If this is true, kyoo will prefer to download the media in the original language of the item.
 MEDIA_PREFER_ORIGINAL_LANGUAGE=false
-# A pattern (regex) to ignore video files.
-LIBRARY_IGNORE_PATTERN=".*/[dD]ownloads?/.*"
+# A pattern (regex) to ignore files.
+LIBRARY_IGNORE_PATTERN=".*/[dD]ownloads?/.*|.*\.(mp3|srt|jpg|jpeg|png|gif|bmp|tiff|svg)$|.*[Tt][Rr][Aa][Ii][Ll][Ee][Rr].*"
 
 # If this is true, new accounts wont have any permissions before you approve them in your admin dashboard.
 REQUIRE_ACCOUNT_VERIFICATION=true
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,8 @@ LIBRARY_ROOT=./video
 # It will automatically be cleaned up on kyoo's startup/shutdown/runtime.
 CACHE_ROOT=/tmp/kyoo_cache
 LIBRARY_LANGUAGES=en
+# If this is true, kyoo will prefer to download the media in the original language of the item.
+MEDIA_PREFER_ORIGINAL_LANGUAGE=false
 # A pattern (regex) to ignore video files.
 LIBRARY_IGNORE_PATTERN=".*/[dD]ownloads?/.*"
 
--- a/scanner/providers/implementations/themoviedatabase.py
+++ b/scanner/providers/implementations/themoviedatabase.py
@@ -149,8 +149,6 @@ class TheMovieDatabase(Provider):
 		Returns:
 			list: A list of images, prioritized by localization, original language, and any available image.
 		"""
-		print(f"Trying to get best image for {item['title']} in {lng}")
-		print(f"All available languageCodes: {map(lambda x: x['iso_639_1'], item['images'][key])}")
 		# Order images by size and vote average
 		item["images"][key] = sorted(
 			item["images"][key],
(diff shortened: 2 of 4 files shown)


zoriya (changes requested):
I am thinking that `prefer_orginial_images` could be a nice **user level** feature for v5 (#597). We might need to change some schema, but since v5 is still pretty early, this is doable.

Having multiples images parsed could also allow us to:
- randomly rotate images (some users might not like that tho idk)
- allow custom selection of images quickly in the front

I'll sit on this a bit and get back during the weekend whether we should merge this now or do something different for v5.

For iso normalization before sending to tmdb this is a bug that we should merge asap. You can either make a separate PR for this we could merge sooner or wait until we decide how we want images handled.

scanner/providers/implementations/themoviedatabase.py
```diff
@@ -254,11 +285,15 @@ async def identify_show(
...
 		async def for_language(lng: str) -> Show:
+			# ISO 639-1 (eg. "en") is required as TMDB does not support fetching images
+			# by ISO 639-2 + ISO 3166-1 alpha-2 (eg. "en-US")
+			lng_iso639_1 = lng.split("-").pop(0)
 			show = await self.get(
 				f"tv/{show_id}",
 				params={
 					"language": lng,
 					"append_to_response": "alternative_titles,videos,credits,keywords,images,external_ids",
+					"include_image_language": f"{lng_iso639_1},null,en",
```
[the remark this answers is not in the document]
felipemarinho97: Yes, you can check this from the docs: https://developer.themoviedb.org/docs/image-languages

> This query (en,null) is looking for all images that match English and those that haven't been set yet (null).

This is just to minimize the chance of getting an empty poster like this.
Screenshot 2024-11-14 at 13 58 03

Note that in my example, using the MR code I still had the poster empty even the [poster exists](https://www.themoviedb.org/movie/1118584-abwab) on TMDB, that's why the MR is still in draft.

scanner/providers/implementations/themoviedatabase.py
```diff
@@ -254,11 +285,15 @@ async def identify_show(
 		languages = self.get_languages()
 
 		async def for_language(lng: str) -> Show:
+			# ISO 639-1 (eg. "en") is required as TMDB does not support fetching images
+			# by ISO 639-2 + ISO 3166-1 alpha-2 (eg. "en-US")
+			lng_iso639_1 = lng.split("-").pop(0)
```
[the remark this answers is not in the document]
felipemarinho97: Good observation, I will refactor the code to use it this way.

felipemarinho97:
>  allow custom selection of images quickly in the front

This is nice, but the preference for original images is something quite niche. If I were you, I would give this as a very low priority to become an user feature to avoid increasing complexity.

I think the feature described in https://github.com/zoriya/Kyoo/issues/87 is already very good and is enough for most people.

felipemarinho97:
I can see 2 ways of fixing https://github.com/zoriya/Kyoo/pull/677#discussion_r1842596897 these empty posters problem.

1. Using a fallback strategy when picking images just like is done in `get_best_image` function.
2. Refactoring the database to allow multiple images as you described, and returning the best available option on-the-fly to the front every request (we would still need a fallback strategy here).

scanner/providers/implementations/themoviedatabase.py
```diff
@@ -138,26 +140,88 @@ def to_studio(self, company: dict[str, Any]) -> Studio:
...
+				item["id"],
+			)
+			images = await self.get(
+				f"{item['media_type']}/{item['id']}/images",
+			)
+			localized_images = sorted(
+				images[key],
+				key=lambda x: x.get("vote_average", 0),
+				reverse=True,
+			)
```
felipemarinho97, line 195: I was able to get 100% of my library items with posters by adding this corner case (of 400, 1 item was a corner case).

I know doing a extra request is bad, and this is also increases the code complexity of this method because it needs to be `async`. If this is too much, I can remove.

scanner/providers/implementations/themoviedatabase.py
```diff
@@ -210,40 +274,27 @@ async def for_language(lng: str) -> Movie:
 				tagline=movie["tagline"] if movie["tagline"] else None,
 				tags=list(map(lambda x: x["name"], movie["keywords"]["keywords"])),
 				overview=movie["overview"],
-				posters=self.get_image(
-					movie["images"]["posters"]
-					+ (
-						[{"file_path": movie["poster_path"]}]
```
zoriya, line 216: do we not need the `poster_path` and things like that from the base api? I added them because it always provided a good fallback if the `images` list was empty.

.env.example
```diff
@@ -10,8 +10,10 @@ LIBRARY_ROOT=./video
 # It will automatically be cleaned up on kyoo's startup/shutdown/runtime.
 CACHE_ROOT=/tmp/kyoo_cache
 LIBRARY_LANGUAGES=en
-# A pattern (regex) to ignore video files.
-LIBRARY_IGNORE_PATTERN=".*/[dD]ownloads?/.*"
+# If this is true, kyoo will prefer to download the media in the original language of the item.
+MEDIA_PREFER_ORIGINAL_LANGUAGE=false
+# A pattern (regex) to ignore files.
+LIBRARY_IGNORE_PATTERN=".*/[dD]ownloads?/.*|.*\.(mp3|srt|jpg|jpeg|png|gif|bmp|tiff|svg)$|.*[Tt][Rr][Aa][Ii][Ll][Ee][Rr].*"
```
zoriya, line 16: you do not need to ignore all those file extensions, we only check for video files (it uses's guessit's mimetype to filter `video/`)

zoriya (commented):
For v5, I already planned the database to contain multiples locales (and image locales). I'll need to find a good way to query those to support `prefer_original` and good fallback.

I left some comments, but I'm pretty happy with the direction/state of the PR. Please request a new review when you feel this is ready, and I'll test & merge.

Thanks for your work on this!

scanner/providers/implementations/themoviedatabase.py
felipemarinho97: Yes!

zoriya: For those cases, is the `poster_path` of the base item empty or does it contain a vaild image?

felipemarinho97: Using `poster_path` seems to be enough, I`ll remove the extra call

.env.example
felipemarinho97: In fact, they are common extensions of extra files that are usually in the same directory as the movie, as I come from jellyfin there are a lot of trickplay files that end up making the scan super slow because guessit is invoked unnecessarily for each thumbnail. I will remove this here, maybe add this as a "tip" in the documentation in another MR.

zoriya:
I'm currently reworking image handling for v5, I haven't forgotten about this dw. It's slowly happening on #680 along with the new api format & multi-language support.

PR #678, opened 2024-11-14
Document scanner's v5 workflow

zoriya:
Part of #597

For v5, we will rework a lot of things in the scanning process. This PR creates a document that details the scanning workflow.

Big changes compared to current version include:
 - The api knows & register even episodes we don't have videos for (or that are not aired).
   - Episodes missing a video will be displayed in the app accordingly.
   - This allows the user to know there's an episode missing and not inadvertently skip one. 
   - We could use this information in admin dashboards.
   - Fetching all episodes was already needed in some cases, this would make them easier.
 - Videos are saved in database before we register their series/movie.
   - We can display them in the interface as `loading` or `fetching-data`.
   - If metadata matching fails, we can still play them.
   - We can manually correct the guess & continue the registration flow.
 - Metadata guesses are saved in database (just guesses based on file name, not external db data)
   - This allows correction to also apply to other episodes of the same series.
   - It gives better information to the user as to why it was guessed this way (or at least it improves the debugging experience for us).

### DIFF
--- a/scanner/README.md
+++ b/scanner/README.md
@@ -1,0 +1,69 @@
+# Scanner
+
+## Workflow (for v5, not current)
+
+In order of action:
+
+ - Scanner gets `/videos` & scan file system to list all new videos
+ - Scanner guesses as much as possible from filename/path ALONE (no external database query).
+   - Format should be:
+     ```json5
+     {
+         path: string,
+         version: number,
+         part: number | null,
+         rendering: sha(path except version & part),
+         guess: {
+             kind: movie | episode | trailer | interview | ...,
+             name: string,
+             year: number | null,
+             season?: number,
+             episode?: number,
+             ...
+          },
+     }
+     ```
+
+     - Apply remaps from lists (AnimeList + thexem). Format is now:
+     ```json5
+     {
+         path: string,
+         version: number,
+         part: number | null,
+         rendering: sha(path except version & part),
+         guess: {
+             kind: movie | episode | trailer | interview | ...,
+             name: string,
+             year: number | null,
+             season?: number,
+             episodes?: number[],
+             absolutes?: number[],
+             externalId: Record<string, {showId, season, number}[]>,
+             remap: {
+                 from: "thexem",
+                 oldSeason: number,
+                 oldEpisodes: number[],
+              },
+             ...
+          },
+     }
+     ```
+ - If kind is episode, try to find the serie's id on kyoo (using the previously fetched data from `/videos`):
+   - if another video in the list of already registered videos has the same `kind`, `name` & `year`, assume it's the same
+   - if a match is found, add to the video's json:
+   ```json5
+   {
+       entries: (uuid | slug | {
+           show: uuid | slug,
+           season: number,
+           episode: number,
+           externalId?: Record<string, {showId, season, number}> // takes priority over season/episode for matching if we have one
+       })[],
+   }
+   ```
+ - Scanner pushes everything to the api in a single post `/videos` call
+ - Api registers every video in the database
+ - For each video without an associated entry, the guess data + the video's id is sent to the Matcher via a queue.
+ - Matcher retrieves metadata from the movie/serie + ALL episodes/seasons (from an external provider)
+ - Matcher pushes every metadata to the api (if there are 1000 episodes but only 1 video, still push the 1000 episodes)
+


scanner/README.md
```diff
@@ -0,0 +1,69 @@
...
+           season: number,
+           episode: number,
+           externalId?: Record<string, {showId, season, number}> // takes priority over season/episode for matching if we have one
+       })[],
+   }
+   ```
+ - Scanner pushes everything to the api in a single post `/videos` call
+ - Api registers every video in the database
+ - For each video without an associated entry, the guess data + the video's id is sent to the Matcher via a queue.
+ - Matcher retrieves metadata from the movie/serie + ALL episodes/seasons (from an external provider)
```
zoriya, line 67: How should we dedup metadata retrieval here? If we have 100 video of `One Piece`, we do not want to fetch every metadata of `One Piece` 100 times.
Should we group videos based on the guessed `name` & `year` fields before sending them as a bulk in the matcher's queue?

scanner/README.md
```diff
@@ -0,0 +1,69 @@
...
+ - Matcher pushes every metadata to the api (if there are 1000 episodes but only 1 video, still push the 1000 episodes)
```
zoriya, line 68: I'm still undecided if this should be done
 - via an api call to post `/series` that would contain all episodes/seasons/...
 - or via a queue of the same data & having the api pull this queue to add items to the database.

v4 uses the api but requires a call per episode + per season + per series so scanning ddos the api.

v5 could benefit from it being an api since external services could want to create movies/series in kyoo since now we can display shows that do not have a video file available.

scanner/README.md
```diff
@@ -0,0 +1,69 @@
+# Scanner
+
+## Workflow (for v5, not current)
+
+In order of action:
+
+ - Scanner gets `/videos` & scan file system to list all new videos
+ - Scanner guesses as much as possible from filename/path ALONE (no external database query).
```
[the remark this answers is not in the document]
zoriya: Yes. The scanner would do tree walk+monitor and for each file run guessit before pushing items.
The matcher would just download/process metadata from online databases & push them to the api.

scanner/README.md
zoriya: Not sure if I understood your first point. What do you mean by `adding multiples items`?
My idea was to have a post /movies & a post /series. A single post would contain every metadata about the item (staff, studio, episodes, seasons...)

I think I'll create an API first because it's the easiest to write & test rn, and it would be cool to have it publicly available for 3rd party apps.
If the API becomes a bottleneck, we could add a queue for the matcher to use.

scanner/README.md
zoriya: The api would receive a bulk add of videos, and would then either:
- create an event per video
- or create an event per video group

This is not stateful, it only affects items sent in the bulk create request.

Adding a cache in the matcher would make things stateful & I feel might be more complex. With a cache we would need to:
- 1 check if item already exist on the api
  - if it does get it's id, link the video to it via an api call
- 2 check if an item of the same show is currently being processed:
  - if yes, wait for it to finish, retrieve it's id and go to step 1
 - 3 download metadata for item
 - 4 push metadata + link video with the metadata

vs just:
- download metadata for item
- push metadata + link videoS with the metadata

scanner/README.md
felipemarinho97: I mean a single endpoint that can handle multiple kinds:

   POST /items `{ "type": "movie|serie|episode" ... other attributes... }`
   
If you decide to have a batch endpoint you can just generalize this:

   POST /items/batch
   ```
   [
     { "type": "movie" ... movie schema... },
     { "type": "serie" ... serie schema... },
     { "type": "episode" ... episode schema... }, # this will depend on serieId maybe
     { "type": "special" ... schema ... },
   ]
   ```
  
  I don't know how is the database schema, so this could be different, but I hope you get the idea

scanner/README.md
felipemarinho97: I'm not sure I understand, wasn't the problem just getting the metadata multiple times?

So, for 100 videos:
- 1. Get metadata -> One Piece
- 2. Get metadata -> One Piece (cached)
and so on..

zoriya:
Merging but the document will evolve with the implementation

PR #679, opened 2024-11-15
Feat/ignore directories support

felipemarinho97:
Features:
- Add `.ignore` support for ignoring directories and subtirectories. 

It is the same funcionalty as `LIBRARY_IGNORE_PATTERN` but it is more manageable for large libraries as it does not require writing a complex regex.

It is also [compatible with Jellyfin](https://jellyfin.org/docs/general/server/media/excluding-directory/), so any user migrating from it will benefit from this.


zoriya (approved):
Sorry I took super long to review this, this is great work!

PR #693, opened 2024-11-27
Update oidc configuration in helm

acelinkio:
This is technically a breaking change, but this chart is still unpublished.  This improvement better aligns with #682

kyoo.oidc_providers is a list of map.  could do as a map of maps, but figured list is easier for others to understand/follow.


zoriya (approved):
thanks!

PR #705, opened 2024-12-09
Add helm chart test and release workflows

joryirving:
Relates to #633 
Creates workflow to test chart using a `kind` cluster. (successful test here: https://github.com/joryirving/Kyoo/actions/runs/12244677078)
Creates workflow to publish chart (example here: https://github.com/joryirving/Kyoo/pkgs/container/helm-charts%2Fkyoo)

Bumps chart version to `0.0.1`

### DIFF
--- a/.github/workflows/helm-release-chart.yml
+++ b/.github/workflows/helm-release-chart.yml
@@ -3,6 +3,7 @@ on:
   push:
     tags:
      - v*
+  workflow_dispatch:
 
 jobs:
   release:
--- a/.github/workflows/helm-release-chart.yml
+++ b/.github/workflows/helm-release-chart.yml
@@ -3,7 +3,6 @@ on:
   push:
     tags:
      - v*
-  workflow_dispatch:
 
 jobs:
   release:
--- a/.github/workflows/helm-release-chart.yml
+++ b/.github/workflows/helm-release-chart.yml
@@ -3,8 +3,6 @@ on:
   push:
     tags:
      - v*
-    paths:
-    - chart/**
   workflow_dispatch:
 
 jobs:
@@ -33,8 +31,11 @@ jobs:
     - name: Package Helm Chart
       working-directory: chart
       run: |
-        helm package .
+        export tag=$(echo ${GITHUB_REF#refs/tags/} | sed 's/^v//')
+        helm package ./chart --version $tag --app-version $tag
         mv *.tgz ../packaged-chart.tgz
+        env:
+          GITHUB_REF: ${{ github.ref }}
 
     - name: Push Helm Chart to GHCR
       run: |
--- a/.github/workflows/helm-release-chart.yml
+++ b/.github/workflows/helm-release-chart.yml
@@ -1,0 +1,43 @@
+name: Release Helm Chart
+on:
+  push:
+    tags:
+     - v*
+    paths:
+    - chart/**
+  workflow_dispatch:
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up Helm
+      uses: azure/setup-helm@v4
+
+    - name: Log in to GHCR
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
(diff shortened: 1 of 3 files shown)


.github/workflows/helm-test-chart.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Lint the Helm chart
+        run: |
+          helm lint ./chart
+
+      - name: Create Kind cluster
+        uses: helm/kind-action@v1
+
+      - name: Install Helm chart in Kind cluster (Dry Run)
+        run: |
+          helm dependency update ./chart
```
zoriya, line 30: why do we run update here?

.github/workflows/helm-release-chart.yml
```diff
@@ -0,0 +1,43 @@
...
+      working-directory: chart
+      run: |
+        helm package .
+        mv *.tgz ../packaged-chart.tgz
+
+    - name: Push Helm Chart to GHCR
+      env:
+        HELM_EXPERIMENTAL_OCI: 1
+      run: |
+        helm push packaged-chart.tgz oci://ghcr.io/${{ github.repository_owner }}/helm-charts
```
zoriya, line 43: shouldn't we name this `kyoo-helm` or something like that?

zoriya (approved):

.github/workflows/helm-test-chart.yml
joryirving: It needs to locally pull down the dependent charts for testing (like meilisearch, rabbit, etc)

.github/workflows/helm-release-chart.yml
joryirving: It appends the package name, so it would be `oci://ghcr.io/zoriya/<path>/kyoo`, it's based on the name in [Chart.yaml](https://github.com/zoriya/Kyoo/blob/master/chart/Chart.yaml#L2).

chart/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.0.0
```
acelinkio, line 18: Could this be swapped to use the tag or release when the workflow is ran?  

Incrementing this on each release is annoying.  `helm package . --version $version` is a bit more straight forward.

.github/workflows/helm-release-chart.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: Package Helm Chart
+      working-directory: chart
+      run: |
+        helm package .
+        mv *.tgz ../packaged-chart.tgz
+
+    - name: Push Helm Chart to GHCR
+      env:
+        HELM_EXPERIMENTAL_OCI: 1
```
acelinkio, line 41: I don't think this env variable is needed anymore with recent versions of helm

https://helm.sh/docs/topics/registries/#enabling-oci-support-prior-to-v380

acelinkio (changes requested):

chart/Chart.yaml
acelinkio: i'm not sure of the actual CI variable name for getting the tag. $version is a placeholder

acelinkio: @zoriya i figured you'd just combine both the Kyoo release and the helm chart release versions together.  Technically the halm chart can be versioned independently.  Just didn't think you'd want to go through the effort.

joryirving: Updated, but I think I need to test it first.

.github/workflows/helm-release-chart.yml
```diff
@@ -0,0 +1,38 @@
+name: Release Helm Chart
+on:
+  push:
```
zoriya, line 3: could we add back a workflow dispatch? i could instantly dispatch a run for `v4.7.0`.

zoriya (approved):

.github/workflows/helm-release-chart.yml
joryirving: No problem!

acelinkio (approved):

PR #708, opened 2024-12-12
feat(helm): Secret creation + rework values structures

qjoly:
Hello there 👋 

The changes proposed in this MR are :
- Re-structure the values so that the chart can be used OOB (without the need to add manifests in `extraObjects`). The modifications are similar to those usually seen in charts (e.g. bitnami) providing a way to either use a pre-existing secret, or generate it from the chart.
- Create volumes within the values

~~- Addition of a new component to the chart to support Ingress natively (without the need to support one-by-one ingress-controllers with their syntaxes).~~ (moved to #710)

Some minor additions: 
- Deletion of unused variables
- Added `strategy` field to define expected behavior during a rollout.
- Remove few `global` sections.

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -1,27 +1,10 @@
 apiVersion: v2
 name: kyoo
 description: A Helm chart for Kubernetes
-
-# A chart can be either an 'application' or a 'library' chart.
-#
-# Application charts are a collection of templates that can be packaged into versioned archives
-# to be deployed.
-#
-# Library charts provide useful utilities or functions for the chart developer. They're included as
-# a dependency of application charts to inject those utilities and functions into the rendering
-# pipeline. Library charts do not define any templates and therefore cannot be deployed.
 type: application
-
-# This is the chart version. This version number should be incremented each time you make changes
-# to the chart and its templates, including the app version.
-# Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.0.0
-
-# This is the version number of the application being deployed. This version number should be
-# incremented each time you make changes to the application. Versions are not expected to
-# follow Semantic Versioning. They should reflect the version the application is using.
-# It is recommended to use it with quotes.
+version: 0.0.1
 appVersion: "4.7.0"
+icon: https://github.com/zoriya/Kyoo/blob/master/icons/icon-256x256.png?raw=true
 
 dependencies:
 - condition: meilisearch.enabled
--- a/chart/README.md
+++ b/chart/README.md
@@ -1,87 +1,49 @@
 # helm chart
 
 # Recomendations
+
 This helm chart includes subcharts for Meilisearch, Postgres, and RabbitMQ.  Those resources should be managed outside of this Helm release.
 
 ## Postgres
+
 Kyoo consists of multiple microservices.  Best practice is for each microservice to use its own database.  Kyoo workloads support best practices or sharing a single postgres database.  Please see the `POSTGRES_SCHEMA` setting for additional information.
 
 Strongly recomended to use a Kubernetes operator for managing Postgres.
 
 ## Storage
-Kyoo currently uses storage volumes for media, backend-storage, and transcoder-storage.  Media content tends to consume a large amount of space and Kubernetes storage interfaces tend to replicate across nodes.  Consider hosting the data outside of Kubernetes or assigning one node to handle storage.
+
+Kyoo currently uses storage volumes for media, backend-storage, and transcoder-storage.  Media content tends to consume a large amount of space and Kubernetes storage interfaces tend to replicate across nodes. Consider hosting the data outside of Kubernetes (e.g by using a networked file system like NFS) or assigning one node to handle storage.
 
 Storage for backend and transcoder will eventually be moved into a datastore application.
 
 # Quickstart
-Below provides an example for deploying Kyoo and its dependencies.  This is a minimalist setup that is not intended for longterm use.  This approach uses a single Postgres instance and initializes mutliple databases.  
 
-```sh
-helm upgrade kyoo . --install --values myvalues.yaml
+Below provides an example for deploying Kyoo and its dependencies.  This is a minimalist setup that is not intended for longterm use.  This approach uses a single Postgres instance and initializes multiple databases.
+
+Clone Kyoo and navigate to the `chart` directory.
+```bash
+git clone https://github.com/zoriya/kyoo.git
+cd kyoo/chart
 ```
-`myvaules.yaml` content
+
+Install the helm chart :
+
+```bash
+helm upgrade kyoo . --install -n kyoo --create-namespace
+```
+
+For production use, it is highly recommended to use a values file to configure unique secrets and settings.
+
 ```yaml
 kyoo:
-  address: https://kyoo.mydomain.com
-meilisearch:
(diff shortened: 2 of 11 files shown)


acelinkio:
I appreciate the time and thought you've invested in suggesting changes.  I want to provide some thoughts and feedback to guide these ideas while keeping the chart flexible and aligned with best practices.

Regarding the use of extraObjects, this feature is intended to demonstrate deploying everything as part of a single Helm release. That said, having Helm directly create secrets is an anti-pattern. Tools like External Secrets are designed specifically for managing secrets securely, separate from the application's lifecycle. Including sensitive information in values.yaml poses a security risk because Kubernetes stores these values in plain text. A better approach is to manage secrets independently to ensure security and flexibility.

On the topic of existingSecrets, its purpose is to allow the chart to rely on pre-created secrets rather than creating new ones. Many Helm charts use this mechanism to give users the flexibility to decide between creating Kubernetes secrets within the chart or using externally managed ones. While I understand the appeal of specifying secret values inline, doing so can introduce unnecessary complexity and make the chart harder to maintain, particularly in deployments involving multiple microservices.  

The use of global values is another key point. Globals enable references throughout the values.yaml file, often to support subcharts.  Subcharts support are best effort to support.  Those resources are best managed external to Kyoo.  Meilisearch in particular requires manual actions with every patch upgrade.

Please note that this is not a small MR like you mentioned https://github.com/zoriya/Kyoo/issues/633#issuecomment-2539604553.  Please break this up into smaller more consumable pieces.  I like the changes that you made for avoiding blank annotations and args.   Being able to specify strategy is also something that should be added.

qjoly:
Thank you for your feedback @acelinkio 🙏 

Unfortunately, we don't have the same vision of how a Helm chart should be managed. I'll try to clarify these points to make sure we understand each other.

> having Helm directly create secrets is an anti-pattern

Helm has to be able to create secrets, it's definitely not an antipattern. Secrets don't always have to be managed by a tool like ESO or SealedSecrets, depending on what they contain (e.g. secrets for accessing a database hosted in Kubernetes: this is totally ok and a very common use).

> Kubernetes stores these values in plain text

Secrets are not encrypted, yes. But they do require a high level of accreditation. Furthermore, tools such as Sealed Secrets, OpenBAO (or Vault) and ESO retrieve tokens from external sources and turn them into secrets. (You always have secrets somewhere).

> While I understand the appeal of specifying secret values inline, doing so can introduce unnecessary complexity and make the chart harder to maintain, particularly in deployments involving multiple microservices

It's a personal opinion, but I don't find it very complex to maintain. Here's a “classic” example found in my code (and in many other Helm charts) : 

```yaml
            - name: KYOO_APIKEYS
              {{- if .Values.kyoo.existingSecret.name }}
              valueFrom:
                secretKeyRef:
                  key: {{ .Values.kyoo.existingSecret.key }}
                  name: {{ .Values.kyoo.existingSecret.name }}
              {{- else }}
              value: {{ .Values.kyoo.apikey }}
              {{- end }}
```

I would add that users are above all looking for an out-of-the-box experience. They certainly want to install Kyoo one-shot, without having to do any additional configuration.

About global values : some were useless (and not 'global' in the real sense). I wanted to simplify the structure of the values to make it clearer.

---

Your feedback is very interesting and it's a debate we need to have to understand the direction in which the changes need to go. I try to apply what I've learned on my own and at work (and I'm still learning every day). To condense my opinion on how a chart should be : 
- It must cover the primary need: installing the application from A to Z (without having to create additional conditions or use extraObjects).
- The chart must be self-sufficient (by default), but it must be possible to integrate it with an existing system (e.g. the chart can deploy a postgres database, but it must also be able to use a database deployed by an operator).
- (in form), the use of helpers.tpl is to be avoided as much as possible (too complex to maintain), there are currently **84 calls** to functions (mainly to name components, which could just be in the form `{{ .Release.Name }}-component_name`)

I sincerely appreciate the time you spent on the review (and sorry for the PR at +500 🙏  I just wanted to make 2-3 changes and got carried away). I've tried to make this message as complete as possible so that we can make decisions about what form the Kyoo chart should take (and so I'll open up other, smaller PRs)

acelinkio:
I think there’s been a bit of a misunderstanding regarding my point about secrets. My concern is about the risks of using Helm (or ArgoCD/Flux/etc.) to apply charts with secrets provided through values.yaml. When this is done, those secrets are inevitably left in plain text somewhere—whether in Git, CI pipelines, or local files—unless additional tooling is introduced.

This isn’t just my personal opinion but aligns with many discussions in the community. Helm is generally seen as a tool that should consume secrets, not create or manage them directly. This topic is especially relevant in GitOps workflows, where handling sensitive data securely is critical. Here’s a related conversation that mirrors my concerns about Helm and user experience in this context: [GitHub issue link](https://github.com/argoproj/argo-cd/issues/5202#issuecomment-2518977683).

On the implementation itself, the current MR forces a Kubernetes secret to be generated, even when existingSecret is used. That was the focus of my earlier comment. However, I’ve since revisited the code to see the switch between between valueFrom and value, and I agree that this is the right way to handle both inline values and existing secrets. Thanks for pointing that out.  

Again, I really do appreciate the work you’ve done here. My intent is to ensure the chart remains flexible and maintainable while addressing user needs  My primary concern is the size and scope of the changes. Large MRs like this one introduce all-or-nothing decisions, which can slow down reviews. Breaking these changes into smaller, focused pieces would make it easier to collaborate, iterate, and get them merged more efficiently.

acelinkio:
I just noticed you also fixed a typo that was happening with extraContainers too.  Your attention to detail is very appreciated.

qjoly:
> On the implementation itself, the current MR forces a Kubernetes secret to be generated, even when existingSecret is used. That was the focus of my earlier comment

Oops, I forgot to add a condition. Thanks for bringing it up 🙏 

I disagree with your point of view on how helm should handle its secrets, even if I do all my deployments in GitOps (like you I guess), it doesn't represent the entire community. But let's keep the main point: the chart is flexible and allows both use cases (including a pre-existing secret and specifying the values in the `values.yaml`).

I opened the PR #710 with the HAProxy setup I suggested. We keep the following subjects here: reworking the value file (proposing the creation of volumes and secrets in the chart)

I have other modifications in mind (notably to factorize deployments / services), the bulk of the work is in this PR so that deployments are independent. I don't think I'll be able to make it any shorter (the changes here are dependent on each other).

Thanks again for your review and your advices !

zoriya:
Thanks for the contribution, we can clearly see that you put some care into it!

I agree with @acelinkio, we'd need to split this up as much as possible to make it easier to review. We could start by a PR to allow secrets to be specified inline. We could go from then by adding small PRs like volume changes, typos...

I think it would be best we discuss a bit more about some changes (like `global`), maybe over discord. I don't want you to do work that would be useless due to misunderstanding!

Btw, I am commenting on this, but @acelinkio is the one who made the chart & the maintainer for it. After discussing a bit, we decided to make it official & I'll add a CODEOWNERS soon!

qjoly:
Thank you for clarifying how this PR should be managed 🙏 

I started to split these modifications into 2 PRs:
- #712 for secret management
- #713 for volume creation

After the review of these two PRs, I could integrate the other changes (unfortunately I can not do it now, it would generate too much conflict or code written to be deleted immediately afterwards).

I close this PRs, thanks again.

Issue #653, opened 2024-10-26
Add prefix every route to kyoo_back


acelinkio:
### Feature description

Following conversation in #633 and https://github.com/zoriya/Kyoo/pull/560#issuecomment-2381556135.  Configuring ingress controllers is a pain.  Every ingress controller implementation has a different method/syntax/approach to effectively doing the same rewrite rule.

To avoid that unneeded complexity, adding a kyoo_back configuration to not require a reverse-proxy rewrite rule would be more straightforward.

PR #680, opened 2024-11-15
v5 api: initial schemas, types & basic routes

zoriya:
This is the first PR for the v5's api (see #597).

Checklist for this PR:
- Database schemas for:
  - [x] shows (movies | series)
  - [x] entries (episode | movie | special | extra)
  - [x] videos
  - [x] seasons
- Api types & examples for:
  - [x] movies
  - [x] series
  - [x] episodes
  - [x] specials
  - [x] extras
  - [x] movies contained in series (called movie-entry here)
  - [x] seasons
- Api setup:
  - [ ] Authorization
  - [x] Error-Problems
  - [x] Clean validations errors (readable)
  - [x] Data seeding for tests (use examples)
  - [x] Clean swagger
- Routes to fetch
  - [x] a movie via id/slug
    - [x] languages
      - [x] query user's requested language
      - [x] ~fallback if we are missing a field~ not planned
      - [ ] allow a `preferOriginal` setting for images
    - [ ] with all translations (if requested)
    - [ ] with all videos
  - [x] all movies
    - [x] w/ arbitrary filters
    - [x] keyset pagination
    - [x] sorts
    - [ ] w/ video (at least if we have one or not, idk if we include all videos at all time)
  - [ ] get a random movie (or sort by random)

Future PRs will be:
- Search (with postgres as the backing db, not meilisearch)
- Images handling
  - [ ] Download images
  - [ ] Scale images (low/medium/high)
  - [ ] Calculate blurhash
  - [ ] Serve images via id
  - [ ] Serve image via /movie/id/poster & co (& support language header here)
- Other get routes (collections, seasons, series, episodes, extra...)
- Mixed routes (shows, news...)
- User specific information (watch-status, history)
- Other metadata (studios, staff...)
- Scanner's workflow (see #678 for the spec)

PR #710, opened 2024-12-13
Add HAProxy route configuration and update ingress settings

qjoly:
Related to #633

To solve the ingress problem for Kyoo, I propose the following modifications to add a new routing component. This component is only present to support url rewriting (without having to conform to the syntax of every existing ingress controller). Depending on Kyoo updates (notably with the proposed issue #653), this component is optional, since its sole purpose is to manage ingress traffic.

I suggest using HAProxy, which is currently the most flexible and easy-to-configure solution (Caddy was also a good candidate, but IMO is less permissive).

Note: This component may not be used by the GatewayAPI configuration (yet to be tested).

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -24,16 +24,9 @@ spec:
           pathType: Prefix
           backend:
             service:
-              name: {{ include "kyoo.front.fullname" . }}
+              name: {{ .Release.Name }}-route
               port:
-                number: 8901
-        - path: "/api"
-          pathType: Prefix
-          backend:
-            service:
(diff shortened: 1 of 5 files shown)


zoriya:
I'd rather we update the different services to make them work without such a complex rewrite (#653). I've been focusing on v5 for a while (see #680) so that's why I haven't tackled it yet, but the change in itself should be pretty straightforward.

This PR is a lot of code to introduce that we would need to maintain even after #653 gets merged (since some users would then rely on the ingress added here.)

We could add an example of ingress/rewrite rules in the documentation to explain that it is needed until #653 is done.

PR #735, opened 2024-12-22
Allow specifying strategy for deployments

acelinkio:
This allows users to specify the update strategy for each deployment.  Setting default for Back and Transcoder to recreate to provide better user experience when using ReadWriteOnce volumes.


acelinkio:
ReadWriteOnce is the default mode that the PVCs this helm charts uses.  Setting to Recreate for the two places its used is important for a smoother user experience.

Eventually we can revisit when the focus is on supporting multiple replicas for back and transcoder

PR #740, opened 2024-12-24
fix fmt sprintf in thumbnails.go

m1k1o:
Forgotten variable, creates this folder:

image


zoriya (approved):
good catch! thanks & merry christmas!

PR #743, opened 2024-12-30
Add basic tv support

scme0:
Add basic tv support by updating Kyoo to use `react-native-tvos` instead of `react-native`.

I've added the exclude so that `expo install --fix` or other upgrade commands don't remove the reference to the tvos package but that means when upgrading, it needs to be upgraded manually.

They way I've been doing it in testing is:
1. removing the exclude
2. running the upgrade
3. manually adding back in the correct equivalent tvos version. `"0.74.5"` -> `"npm:react-native-tvos@0.74.5-0"`

I also added the config-tv plugin which helps configure the android and ios resources for tv during build.

I've purposefully not upgraded any versions and use the equivalent tvos version as I've found upgrading doesn't seem to work for me. This needs more investigation.

This doesn't mean tv is supported as there are several immediate issues which make it practically unusable. I'll be creating PRs to fix things as I find and solve them.

I've tested the normal mobile build (without `EXPO_TV=1`) on my phone for a few days and everything seems to work fine. I think it's pretty safe but happy for someone else to give it a go to be sure.


zoriya:
Thanks for starting this! I'll try to release v4.7.1 soon & merge this shortly after!

PR #747, opened 2025-01-02
Add prefix to back/auth/transcoder

### LINKED ISSUE #653: Add prefix every route to kyoo_back

zoriya:
This is a breaking change for downstream transcoder users that uses the `GOCODER_PREFIX` env var (ping @Arthi-chaud)

Closes #653 (will still need to add this option for v5's api, but no need to keep the issue around)

### DIFF
--- a/auth/.env.example
+++ b/auth/.env.example
@@ -1,6 +1,9 @@
 # vi: ft=sh
 # shellcheck disable=SC2034
 
+# http route prefix (will listen to $KEIBI_PREFIX/users for example)
+KEIBI_PREFIX=""
+
 # Database things
 POSTGRES_USER=kyoo
 POSTGRES_PASSWORD=password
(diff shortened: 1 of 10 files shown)


zoriya:
ping @acelinkio for the chart update

zoriya:
(i also reverted #700 since https://github.com/meilisearch/meilisearch-dotnet/issues/601 is an issue on the latest meilisearch release)

PR #749, opened 2025-01-02
path update for helm charts

acelinkio:
relates to https://github.com/zoriya/Kyoo/pull/747  updates the helm chart to specify paths and avoid the need for rewrite rules.

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -21,7 +21,7 @@ version: 0.0.0
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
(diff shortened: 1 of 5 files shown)


acelinkio:
@zoriya , i can change the target for this MR to your feature branch if you'd like.  note that this will pull in some changes from master branch.

I don't exactly follow why the `/video` path needs to be specified in the transcoder url.  Eventually I'll get around to building/publishing the containers to test more locally.

zoriya:
Some routes of the transcoder return URIs that should be reachable from the front (that was the previous use case of the `GOCODER_PREFIX` env var.) The c# back proxies this api under `/video` so we need to specify this `prefix` value (and it changes the listen route to include this prefix.)
On v5, the transcoder will be accessed directly without the proxy so it will be less weird x)

Issue #753, opened 2025-01-03
Some subtitles with Hearing Impaired tag (HI) are detected as Hindi subtitles.


felipemarinho97:
### Kyoo's version

4.7.0

### What happened?

<img width="462" alt="Screenshot 2025-01-03 at 15 36 22" src="https://github.com/user-attachments/assets/61893860-d2dc-40c0-adec-c3d60357c42e" />

All subtitles above are "Portuguese", the last one is for Hearing Impaired people. But is showing as Hindi.

Issue #588, opened 2024-08-08
Automatic subtitle selection


JeWe37:
### Feature description

It would make sense to automatically select subtitles for videos based on user preferences. 

### Examples 

For instance a user, that only knows English and German, will want English or German non-forced subtitles if they are watching a Spanish show. Similarly, forced subtitles should always be automatically selected if their language matches the audio language and the user knows that language. Technically speaking default subtitles should always be turned on, but I find the tagging for those to be generally too inconsistent as to be worth considering. If two subtitles are of the same preference, some ranking should be done, perhaps based on the subtitle codec, preferring e.g. typically higher quality .ass subtitles.

### References

For an approach for how to do the sorting, see for instance jellyfin/jellyfin#11072.

zoriya:
There is already an option for this:
![image](https://github.com/user-attachments/assets/dbbeff54-961e-43aa-8afe-2c330dc87fca)

Do you want improvements over this option? Right now, there is no CC/SDH special handling.

Forced subtitles are not automatically enabled when you select the corresponding audio, it could make a nice addition.

JeWe37:
Ah, I missed that one. Then indeed, let's change this to "Automatic subtitle selection improvements". Most of the ideas, SDH, handling multiple known languages, forced subtitle handling, as well as inly activating subtitles for unknown languages still apply.

zoriya:
This is implemented on master, if someone has an idea of another improvement please open another issue.

PR #754, opened 2025-01-03
Feat/Hearing Impaired flag support

felipemarinho97:
Closes https://github.com/zoriya/Kyoo/issues/753

### DIFF
--- a/transcoder/src/subtitles.go
+++ b/transcoder/src/subtitles.go
@@ -69,9 +69,14 @@ outer:
 				// Handle Hindi (hi) collision with Hearing Impaired (hi):
 				// "hi" by itself means a language code, but when combined with other lang flags it means Hearing Impaired.
 				// In case Hindi was not detected before, but "hi" is present, assume it is Hindi.
-				if sub.Language == nil && strings.Count(strings.Join(flags, " "), "hi") > 1 {
-					langStr := language.Hindi.String()
-					sub.Language = &langStr
+				if sub.Language == nil {
+					for _, flag := range flags {
+						if flag == "hi" {
+							langStr := language.Hindi.String()
+							sub.Language = &langStr
+							break
+						}
+					}
 				}
 
 				mi.Subtitles = append(mi.Subtitles, sub)
--- a/transcoder/src/subtitles.go
+++ b/transcoder/src/subtitles.go
@@ -58,15 +58,10 @@ outer:
 					default:
 						lang, err := language.Parse(flag)
 						if err == nil && lang != language.Und {
-							lang := lang.String()
-							sub.Language = &lang
+							langStr := lang.String()
+							sub.Language = &langStr
 						} else {
 							sub.Title = &flag
-						}
-						// if language is Hindi, remove hearing impaired flag
-						// ambiguity is solved but as a result we do not support HI for Hindi
-						if lang == language.Hindi {
-							sub.IsHearingImpaired = false
 						}
 					}
 				}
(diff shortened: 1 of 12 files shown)


zoriya (commented):
Thanks that's a great addition!

We might also want to add the `CC` info on the subtitle selection menu.

zoriya:
This will give more options for #588 also!

felipemarinho97:
> Thanks that's a great addition!
> 
> We might also want to add the `CC` info on the subtitle selection menu.

Sure! added all the additional info in the subtitle menu.
Screenshot 2025-01-03 at 18 24 35


Reference:
```
-rw-r--r-- 1 user user     45989 Jan  3 15:59 Blade II (2002) 720p BRRip x264 -YTS.hi.hi.srt
-rw-r--r-- 1 user user     45989 Jan  3 16:17 Blade II (2002) 720p BRRip x264 -YTS.hin.hi.srt
-rw-r--r-- 1 user user     45989 Jan  3 16:15 Blade II (2002) 720p BRRip x264 -YTS.hin.srt
-rw-r--r-- 1 user user     45989 Jan  3 15:59 Blade II (2002) 720p BRRip x264 -YTS.hi.srt
-rw-r--r-- 1 user user 523817836 Sep  9  2023 Blade II (2002) 720p BRRip x264 -YTS.mkv
-rw-r--r-- 1 user user     45989 Jan  3 15:59 Blade II (2002) 720p BRRip x264 -YTS.pt-BR.hi.srt
-rw-r--r-- 1 user user     45989 Oct 23 18:56 Blade II (2002) 720p BRRip x264 -YTS.pt-BR.srt
```

transcoder/src/subtitles.go
```diff
@@ -53,16 +53,36 @@ outer:
...
+					for _, flag := range flags {
+						if strings.EqualFold(flag, "hi") {
+							hiCount++
+							langStr := language.Hindi.String()
+							sub.Language = &langStr
+						}
+					}
+					if hiCount == 1 {
+						sub.IsHearingImpaired = false
+					}
```
zoriya, line 83: Is there a reason for `EqualFold` here?
I'd rather:

```suggestion
					hiCount := Count(flags, "hi")
					if hiCount > 0 {
						langStr := language.Hindi.String()
						sub.Language = &langStr
					}
					if hiCount == 1 {
						sub.IsHearingImpaired = false
					}
```

with Count being defined in `utils.go` as:

```go
func Counts[S []E, E comparable](s S, e E) int {
	var n int
	for _, v := range s {
		if v == e {
			n++
		}
	}
	return n
}
```

zoriya (approved):
Left two minor comments, other than that it's good to merge. Thanks for your work!

transcoder/src/subtitles.go
felipemarinho97: The code before is using `strings.ToLower(flag)`. I liked your suggestion, but it will not work if the "HI" flag is uppercase.

zoriya: Good catch, we could switch on `flag` instead of `strings.ToLower(flag)` and change

```go
flags := separator.Split(match[len(base_path):], -1)
```
with

```go
flags_str := strings.ToLower(match[len(base_path):])
flags := separator.Split(flags_str, -1)
```

felipemarinho97: Good, this will work

zoriya (approved):

PR #755, opened 2025-01-03
Handle non resize transcode

### LINKED ISSUE #626: Quality detection when aspect ratio != 16/9
### Kyoo's version

_No response_

### What happened?

We detect the video's quality using video height, this yield one quality lower than expected when aspect ratio is not 16/9.
This is only applicable for video files that can't be natively played.

zoriya:
Fix #626

(still needs a bit of testing but should be fine)


zoriya:
This is indeed fixed
image

PR #763, opened 2025-01-05
Migrate to expo only (no next) PART 1

zoriya:
This is a draft keeping track of the migration to [one](https://onestack.dev/). This will allow us to delete the expo+nextjs setup and only use a single router + root package.json.

Blockers:
- ~Can't access `<head>` nor add css during the ssr pipeline. Yoshiki can't be used as is.~
- There doesn't seem to be loaders only triggered during ssr, idk how we could prefetch react-query during ssr only.
- ~Babel config is not exposed so even if we migrate to unistyle, we wouldn't be able to set it up.~


zoriya:
This is diverging too much, i'll merge as is and continue to work on master with smaller PRs

zoriya:
CI fails due to timeout, skipping

PR #843, opened 2025-03-11
Add watch status in entries, movies & series

zoriya:
Also create the `history` & `watchlist` tables.

This PR is mostly made to validate the sql schema, we'll need to add auth to the api before it can be merged.


zoriya:
We should probably return the `progress` in `/movies` (& `/shows`) also (maybe not the time tho, idk)

PR #857, opened 2025-03-23
Integerate the auth service to the api

zoriya:
This uses the [ForwardAuth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) middleware of traefik to implement phantom tokens for the api.

Every request with an `Authorization` header is forwarded to the `http://auth/jwt` route which convert the opaque token to a jwt. This jwt is then sent to the api that can read it to retrieve user's information.

Todo:
 - [x] Find a way to actually retrieve the jwt, it's not forwarded rn (but 401/403 from the auth are returned to the client without going through the api)
 - [x] Access the user in a simple way from api routes (probably an elysia macro)
 - [ ] Proctect every route with a set of permissions/claims
 - [ ] Find a way to test this in CI

In other PRs: 
- [ ] Cleanup auth service (auth handling assumes it's already gone through the `/jwt` route which is not possible)
- [ ] Merge the auth & api swaggers
- [ ] Add 401/403 status to every route of the api
- [ ] Make a good guest handling (i think we can generate temp sessions in the auth service and the api should be able to proceed like every connection is a valid user)
- [ ] Add `correlations-id` in every request
- [ ] Handle user deletions (the easiest would be to add a foreign key across db schemas, but that's probably a bad idea)

cc @acelinkio, idk how to handle this in k8s. this would probably require the chart to bundle a gateway, no?


acelinkio:
Hey.  I'll need to do some reading to see what tools support Phantom Tokens to choose the best course.  Here are my initial thoughts:

**Option 1:**
Introduce standalone traefik installation to mirror what docker-compose based setup looks like.  

Traefik -> Kyoo Workloads

pros: keeps networking minimal hops
cons: have to manually add tls cert. must use separate ip address from other ingress/httproutes

**Option 2:**
Similar to option 1, Introduce standalone traefik installation to mirror what docker-compose based setup looks like.  Add ingress/httproute ahead of.

ingress/httproute -> Traefik -> Kyoo Workloads

pros: terminates tls inside of ingress/httproute
cons: extra hop

**Option 3:**
Investigate if common ingress/gateway controllers are working on support for phantom tokens.  
* ingress: nginx-ingress
* ingress: haproxy
* ingress: traefik
* gateway: cilium https://gateway-api.sigs.k8s.io/implementations/#cilium
* gateway: envoygateway https://gateway-api.sigs.k8s.io/implementations/#traefik-proxy
* gateway: traefik https://gateway-api.sigs.k8s.io/implementations/#traefik-proxy

then add support/documentation for those that do

zoriya:
oww that looks like a pain to do, kinda sorry about that x)

for nginx I found https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#external-authentication

acelinkio:
MIddleware is always a pain to manage. Let me know when a working docker-compose based example is available. I can start experimenting after that

zoriya:
The middleware/auth flow now works! (unlogged in users are still allowed for now)

I'll start by cleaning the auth service and slowly make PRs to tackles remaining tasks.

PR #896, opened 2025-04-20
[v5] Added support for storing images in S3-compatible object storage

solidDoWant:
As discussed on Discord [here](https://discord.com/channels/1216460898139635753/1216460898609401870/1362975074713469058), this adds support for storing images in S3-compatible object storage for the v5 implementation.

The benefit of using S3 here is that when multiple replicas of the service are running, they will now all have access to the same set of images. Returned values for `/images/:id` should now be consistent, regardless of which replica serves a response. In addition, images should only be downloaded and stored once for the entire service, instead of once per replica.

The environment variables that should be used to configure the S3 client are documented here: https://bun.sh/docs/api/s3#credentials

### DIFF
--- a/api/src/controllers/images.ts
+++ b/api/src/controllers/images.ts
@@ -369,9 +369,9 @@ export async function isCached(
 		let lastModified: Date | undefined;
 		const stat = await getFile(filePath).stat();
 		try {
-			if (stat as S3Stats) {
+			if ((stat as S3Stats).lastModified) {
 				lastModified = (stat as S3Stats).lastModified;
-			} else if (stat as Stats) {
+			} else if ((stat as Stats).mtime) {
 				lastModified = (stat as Stats).mtime;
 			}
 		} catch {
--- a/api/src/controllers/images.ts
+++ b/api/src/controllers/images.ts
@@ -1,7 +1,7 @@
+import type { Stats } from "node:fs";
 import type { BunFile, S3File, S3Stats } from "bun";
 import { type SQL, and, eq, sql } from "drizzle-orm";
 import Elysia, { type Context, t } from "elysia";
-import type { Stats } from "node:fs";
 import { prefix } from "~/base";
 import { db } from "~/db";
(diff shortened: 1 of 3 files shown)


api/src/controllers/images.ts
```diff
@@ -366,8 +367,13 @@ export async function isCached(
 	if (headers["if-modified-since"]) {
 		const ifModifiedSince = headers["if-modified-since"];
 		let lastModified: Date | undefined;
+		const stat = await getFile(filePath).stat();
 		try {
-			lastModified = (await stat(filePath)).mtime;
+			if (stat as S3Stats) {
```
zoriya, line 372: I'm pretty sure this won't work. typescript types are just syntactic sugar and as won't return undefined if it's the wrong type.

you could do something like

```typescript
if ("lastModified" in stat)
```

instead

api/src/controllers/seed/images.ts
```diff
@@ -1,18 +1,16 @@
...
 import { version } from "package.json";
 import type { PoolClient } from "pg";
 import sharp from "sharp";
 import { type Transaction, db } from "~/db";
 import { mqueue } from "~/db/schema/mqueue";
 import type { Image } from "~/models/utils";
+import { getFile } from "~/utils";
 
 export const imageDir = process.env.IMAGES_PATH ?? "./images";
-await mkdir(imageDir, { recursive: true });
```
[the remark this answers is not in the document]
solidDoWant: I don't think this is the case. From the `Bun.write` docs:

> If `destination`'s directory does not exist, it will be created by default.

I'll add the mkdir call back if you like - I just don't think that it is needed here.

api/src/controllers/images.ts
solidDoWant: Thanks! I was not aware of this. I read up a little and I _think_ I'm doing it the idiomatic way now. PTAL and I'll update again if I'm wrong.

PR #899, opened 2025-04-22
Support standard libpq environment variables for the API

solidDoWant:
The implements standard libpq environment variables documented [here](https://www.postgresql.org/docs/current/libpq-envars.html). The current environment variables are now supported:
* `POSTGRES_URL` - nonstandard. Used to set the connection string. Be aware that configuration values provided via the connection string are limited to what the Postgres driver's parser logic supports.
* `PGHOSTADDR`
* `PGHOST` - already supported, but renamed while retaining support for the current env var name.
* `PGPORT` - already supported, but renamed while retaining support for the current env var name.
* `PGDATABASE` - already supported, but renamed while retaining support for the current env var name.
* `PGUSER` - already supported, but renamed while retaining support for the current env var name.
* `PGPASSWORD` - already supported, but renamed while retaining support for the current env var name.
* `PGOPTIONS`
* `PGCONNECT_TIMEOUT`
* `PGPASSFILE`
* `PGSSLMODE` - This differs from the spec a bit. If anything other than `disable` is specified, only TLS will be attempted. In other words, `allow` and `prefer` will not attempt a non-TLS connection. Supporting this would require quite a bit of extra logic that attempts to create connections with and without TLS, and probably isn't desired.
* `PGSSLROOTCERT` - This differs from the spec when `system` is specified. It looks like Bun does not support `tls.getCACertificates`, so when a value of `system` is specified, the bundled root CA certs are used instead of the system certs.
* `PGSSLCERT`
* `PGSSLKEY`
* `PGSSLSNI`
* `PGSSLMINPROTOCOLVERSION`
* `PGSSLMAXPROTOCOLVERSION`

With this PR, the following postgres features are now supported (when configured):
* Connection via TLS with a Postgres serving cert that is not publicly trusted
* Postgres authentication via client certificate (mTLS), removing the need for long-lived passwords
* Connecting to a postgres server without DNS resolution
* Lowering or raising the connection timeout, allowing for quickly determining when the DB is offline, or allowing more time for slower networks
* Loading a password via a file instead of an env var
* Setting connection options
* Configuring the postgres connection via a single env var (with some limitations)

### DIFF
--- a/api/.env.example
+++ b/api/.env.example
@@ -14,8 +14,18 @@ AUTH_SERVER=http://auth:4568
 
 IMAGES_PATH=./images
 
-POSTGRES_USER=kyoo
-POSTGRES_PASSWORD=password
-POSTGRES_DB=kyooDB
-POSTGRES_SERVER=postgres
-POSTGRES_PORT=5432
+# It is recommended to use the below PG environment variables when possible.
+POSTGRES_URL=postgres://user:password@hostname:port/dbname?sslmode=verify-full&sslrootcert=/path/to/server.crt&sslcert=/path/to/client.crt&sslkey=/path/to/client.key
+# The behavior of the below variables match what is documented here:
+# https://www.postgresql.org/docs/current/libpq-envars.html
+PGUSER=kyoo
+PGPASSWORD=password
+PGDB=kyooDB
+PGSERVER=postgres
+PGPORT=5432
+PGOPTIONS=-c search_path=kyoo,public
+PGPASSFILE=/my/password # Takes precedence over PGPASSWORD. New line characters are not trimmed.
+PGSSLMODE=verify-full
+PGSSLROOTCERT=/my/serving.crt
+PGSSLCERT=/my/client.crt
+PGSSLKEY=/my/client.key
--- a/api/.env.example
+++ b/api/.env.example
@@ -15,7 +15,7 @@ AUTH_SERVER=http://auth:4568
 IMAGES_PATH=./images
 
 # It is recommended to use the below PG environment variables when possible.
-POSTGRES_URL=postgres://user:password@hostname:port/dbname?sslmode=verify-full&sslrootcert=/path/to/server.crt&sslcert=/path/to/client.crt&sslkey=/path/to/client.key
+# POSTGRES_URL=postgres://user:password@hostname:port/dbname?sslmode=verify-full&sslrootcert=/path/to/server.crt&sslcert=/path/to/client.crt&sslkey=/path/to/client.key
 # The behavior of the below variables match what is documented here:
 # https://www.postgresql.org/docs/current/libpq-envars.html
 PGUSER=kyoo
@@ -23,9 +23,9 @@ PGPASSWORD=password
 PGDB=kyooDB
 PGSERVER=postgres
 PGPORT=5432
-PGOPTIONS=-c search_path=kyoo,public
-PGPASSFILE=/my/password # Takes precedence over PGPASSWORD. New line characters are not trimmed.
-PGSSLMODE=verify-full
-PGSSLROOTCERT=/my/serving.crt
-PGSSLCERT=/my/client.crt
-PGSSLKEY=/my/client.key
+# PGOPTIONS=-c search_path=kyoo,public
+# PGPASSFILE=/my/password # Takes precedence over PGPASSWORD. New line characters are not trimmed.
+# PGSSLMODE=verify-full
+# PGSSLROOTCERT=/my/serving.crt
+# PGSSLCERT=/my/client.crt
+# PGSSLKEY=/my/client.key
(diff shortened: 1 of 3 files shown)


solidDoWant:
I need to do some testing prior to merge, but wanted to open this first to make sure that this is in line with the project's goals

api/src/db/index.ts
```diff
@@ -1,19 +1,162 @@
...
+	}
+
+	config.ssl = ssl;
+	return config;
+}
+
 export const db = drizzle({
 	schema,
-	connection: dbConfig,
+	connection: await getPostgresConfig(),
```
zoriya, line 159: We'd want to store the config as a const value to not await this 3 times when calling the `migrate` function bellow

api/src/db/index.ts
```diff
@@ -1,19 +1,162 @@
...
+	// If the config is a hostname and the host address is set, use a custom lookup function
+	if (net.isIP(config.host ?? "") === 0 && process.env.PGHOSTADDR) {
+		const ipVersion = net.isIP(process.env.PGHOSTADDR);
+		if (ipVersion === 0) {
+			throw new Error(
+				`PGHOSTADDR is not a valid IP address: ${process.env.PGHOSTADDR}`,
+			);
+		}
+
+		(config.ssl as ConnectionOptions).lookup = (
```
zoriya, line 49: why do we want/need a custom dns lookup?

api/src/db/index.ts
```diff
@@ -1,19 +1,162 @@
...
+	// that non-SSL connection options under `ssl` (which is incorrectly named) cannot be set unless SSL is enabled.
+	if (!process.env.PGSSLMODE || process.env.PGSSLMODE === "disable")
+		return config;
+
+	// Despite this field's name, it is used to configure everything below the application layer.
+	const ssl: ConnectionOptions = {
+		timeout:
+			(process.env.PGCONNECT_TIMEOUT &&
+				Number(process.env.PGCONNECT_TIMEOUT)) ||
+			undefined,
```
zoriya, line 35:
```suggestion	Number(process.env.PGCONNECT_TIMEOUT) || undefined,
```

api/src/db/index.ts
solidDoWant: So the "host" field is used in three places when connecting to a postgres server: DNS resolution (not done if the host field is an IP address), TLS SNI extension so the server knows which cert to provide, and hostname verification of the certificate provided by the server.

There are a few reason why somebody might want to be able to override this:
* Specifying the IP address to connect to can remove the need for DNS lookups entirely. In some environments this can be important, either because DNS resolution is unavailable, or because there is a concern about exfiltrating data (DNS requests can be used to send data out, even if the service can only talk to a recursive resolver). Setting the PGHOSTADDR var and the PGHOST allows for this while also enabling SNI and hostname verification.
* It can be used to verify the server's certificate when the server provides (and is expected to provide) a certificate with a different domain name. For example, you can use this when doing local development if you're using something like `kubectl port-forward` to access a database in-cluster. In this case, the database would probably have certificate with the service name (say `postgres.media.svc.cluster.local`), but because the connection is being proxied via `kubectl port-forward`, the client need to connect to a localhost address. With this option, the address could be set to `127.0.0.1`, while verifying that the certificate is for `postgres.media.svc.cluster.local`

    I've also heard (but not verified myself) that some cloud services like GCP use a generic cert for managed postgres instances. This is another place that it could be useful.

I'm happy to remove this if you like. It could be useful for local dev in some cases, and maybe other users would find it useful, but I personally don't need it.

zoriya: nah that's fine, thanks for the explanation!

zoriya:
You should update `.env.example` with a link to the doc/list of all values supported

solidDoWant:
@zoriya this should be ready for merge if it meets your approval. I've manually tested all options.

I've also removed the following lesser-used options. It seems like the postgres driver ignores them for some reason:
- `PGHOSTADDR`
- `PGCONNECT_TIMEOUT`
- `PGSSLSNI`
- `PGSSLMINPROTOCOLVERSION`
- `PGSSLMAXPROTOCOLVERSION`

The remaining options should still be sufficient for TLS connections + TLS auth.

zoriya (approved):

api/.env.example
```diff
@@ -14,8 +14,18 @@ AUTH_SERVER=http://auth:4568
...
-POSTGRES_USER=kyoo
-POSTGRES_PASSWORD=password
-POSTGRES_DB=kyooDB
-POSTGRES_SERVER=postgres
-POSTGRES_PORT=5432
+# It is recommended to use the below PG environment variables when possible.
+POSTGRES_URL=postgres://user:password@hostname:port/dbname?sslmode=verify-full&sslrootcert=/path/to/server.crt&sslcert=/path/to/client.crt&sslkey=/path/to/client.key
+# The behavior of the below variables match what is documented here:
+# https://www.postgresql.org/docs/current/libpq-envars.html
+PGUSER=kyoo
```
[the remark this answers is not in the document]
solidDoWant: Sure - done

zoriya (approved):

PR #901, opened 2025-04-24
Support standard libpq environment variables for Auth service

solidDoWant:
Follow up to https://github.com/zoriya/Kyoo/pull/899. This uses the pgx library to parse a connection string and env vars. The pgx library supports enough env vars for TLS connections, and TLS auth (on top of the currently supported vars).


zoriya (approved):
this is really nice, we should also migrate the transcoder to pgx soonish

PR #904, opened 2025-04-25
Add support for RabbitMQ connection string to Backend service

solidDoWant:
This adds support for specifying the rabbitmq connection details via a connection string.

Unfortunately, while the official rabbitmq client supports parsing basic connection string options, it does not support parsing anything that is provided via query parameters. As a result, this has been added to the application (rather than calling a library function).

Backwards compatibility has been preserved.

~~I need to do some testing prior to merge. I'm going to work on adding this support to other the services, and then I'll manually test them all at once.~~

### DIFF
--- a/back/src/Kyoo.RabbitMq/RabbitMqModule.cs
+++ b/back/src/Kyoo.RabbitMq/RabbitMqModule.cs
@@ -96,7 +96,8 @@ public static class RabbitMqModule
 		if (!string.IsNullOrEmpty(caCertFile))
 		{
 			// Load the cert once at startup instead of on every connection.
-			X509Certificate2 rootCA = new(caCertFile);
+			X509Certificate2Collection rootCACollection = [];
+			rootCACollection.ImportFromPemFile(caCertFile);
 
 			// This is a custom validator that obeys the set SslPolicyErrors, while also using the CA cert specified in the query string.
 			factory.Ssl.CertificateValidationCallback = (
@@ -144,7 +145,7 @@ public static class RabbitMqModule
 				);
 				chain.ChainPolicy.CustomTrustStore.Clear();
 				chain.ChainPolicy.TrustMode = X509ChainTrustMode.CustomRootTrust;
-				chain.ChainPolicy.CustomTrustStore.Add(rootCA);
+				chain.ChainPolicy.CustomTrustStore.AddRange(rootCACollection);
 
 				return chain.Build(new X509Certificate2(certificate));
 			};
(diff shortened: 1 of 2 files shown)


solidDoWant:
This is working. Tested via `RABBITMQ_URL="amqps://localhost:5671?cacertfile=/tmp/serving-cert-1/owned-cert.pem&certfile=/tmp/ca-cert-1/guest.crt&keyfile=/tmp/ca-cert-1/guest.key&auth_mechanism=external"`, as well as pre-existing env vars

back/src/Kyoo.RabbitMq/RabbitMqModule.cs
```diff
@@ -30,18 +33,216 @@ public static void ConfigureRabbitMq(this WebApplicationBuilder builder)
...
+		else if (factory.Port == 0)
+			factory.Port = 5672;
+	}
+
+	private static string _GetNonEmptyString(params string?[] values)
+	{
+		foreach (var value in values)
+			if (!string.IsNullOrEmpty(value))
+				return value;
+		return string.Empty;
```
zoriya, line 246:
```suggestion
		return values.FirstOrDefault(string.IsNullOrEmpty) ?? string.Empty;
```

back/src/Kyoo.RabbitMq/RabbitMqModule.cs
```diff
@@ -30,18 +33,216 @@ public static void ConfigureRabbitMq(this WebApplicationBuilder builder)
...
+		factory.HostName = _GetNonEmptyString(
+			configuration.GetValue<string>("RABBITMQ_HOST"),
+			factory.HostName,
+			"rabbitmq"
+		);
+		var port = configuration.GetValue<int?>("RABBITMQ_PORT");
+		if (port != null)
+			factory.Port = port.Value;
+		else if (factory.Port == 0)
+			factory.Port = 5672;
```
zoriya, line 238:
```suggestion
		factory.Port = configuration.GetValue<int>("RABBITMQ_PORT", 5672);
```

back/src/Kyoo.RabbitMq/RabbitMqModule.cs
```diff
@@ -30,18 +33,216 @@ public static void ConfigureRabbitMq(this WebApplicationBuilder builder)
...
+		factory.Uri = new Uri(connectionString);
+
+		// Support query parameters defined here:
+		// https://www.rabbitmq.com/docs/uri-query-parameters
+		Dictionary<string, Microsoft.Extensions.Primitives.StringValues> queryParameters =
+			QueryHelpers.ParseQuery(factory.Uri.Query);
+
+		queryParameters.TryGetValue(
+			"heartbeat",
+			out Microsoft.Extensions.Primitives.StringValues heartbeats
```
zoriya, line 69: Can't `Microsoft.Extensions.Primitives` be added as a `using`?

zoriya (approved):
Just for potential future readers: rabbitmq will go away with v5 (as well as the current back implementation)

Thanks for this PR, it looks like a lot of work ngl

zoriya (approved):

PR #905, opened 2025-04-25
Add support for RabbitMQ connection string to the Autosync service

solidDoWant:
Follow-up to https://github.com/zoriya/Kyoo/pull/904. ~~As with the other PR, I'll test this after I've added this support to all services.~~

### DIFF
--- a/autosync/.env.example
+++ b/autosync/.env.example
@@ -1,0 +1,12 @@
+# vi: ft=sh
+# shellcheck disable=SC2034
+
+# RabbitMQ settings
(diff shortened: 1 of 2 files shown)


solidDoWant:
Tested this via `RABBITMQ_URL="amqps://localhost:5671/?cafile=/tmp/serving-cert-1/owned-cert.pem&certfile=/tmp/ca-cert-1/guest.crt&keyfile=/tmp/ca-cert-1/guest.key&auth=external"` and it's working great.

zoriya (approved):

PR #908, opened 2025-04-26
Update workflows to work on forks

solidDoWant:
There are a handful of workflow steps that cannot be ran on forks (at least easily). This selectively enables some workflow logic only if the required secrets are present. It also removes the hard-coded "zoriya" prefix from image names so that images able to be pushed to forked repos when a release is cut.

Here is a release I cut on my fork with these changes: https://github.com/solidDoWant/Kyoo/releases/tag/v4.8.0-dev.4


zoriya (approved):
Thanks!

PR #909, opened 2025-04-27
Fix Backend service crash when RabbitMQ queues are predeclared

solidDoWant:
If queues or exchanges are predeclared with different options (such as `durable = true` or `type = cluster`), the backend service will crash on startup. This checks if the queues and exchanges exist prior to re-declaring them, allowing users to pre-configure these  with non-default options.

### DIFF
--- a/back/src/Kyoo.RabbitMq/RabbitProducer.cs
+++ b/back/src/Kyoo.RabbitMq/RabbitProducer.cs
@@ -30,7 +30,8 @@ public class RabbitProducer
 	{
 		_channel = rabbitConnection.CreateModel();
 
-		_channel.ExchangeDeclare("events.resource", ExchangeType.Topic);
+		if (!doesExchangeExist(rabbitConnection, "events.resource"))
+			_channel.ExchangeDeclare("events.resource", ExchangeType.Topic);
 		_ListenResourceEvents<Collection>("events.resource");
 		_ListenResourceEvents<Movie>("events.resource");
 		_ListenResourceEvents<Show>("events.resource");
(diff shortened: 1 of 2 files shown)


back/src/Kyoo.RabbitMq/RabbitProducer.cs
```diff
@@ -30,7 +30,8 @@ public RabbitProducer(IConnection rabbitConnection)
 	{
 		_channel = rabbitConnection.CreateModel();
 
-		_channel.ExchangeDeclare("events.resource", ExchangeType.Topic);
+		if (!doesExchangeExist(rabbitConnection, "events.resource"))
+			_channel.ExchangeDeclare("events.resource", ExchangeType.Topic);
```
zoriya, line 34: Shouldn't we instead declare and ignore an exception if it is already configured? i think the current implementation can crash due to race conditions.

PS: i'm also fine with merging it as is, even if it crashes due to race migration, the app should restart instantly and use the existing channel.

solidDoWant: > Shouldn't we instead declare and ignore an exception if it is already configured?

I looked into this approach, but from what I can tell, the RabbitMQ library throws the same exception for most errors returned by the server. This makes it really difficult to tell say, a permissions error and a queue conflict error apart.

You are absolutely right, this is a race condition. Would you prefer that I try to parse the exception message (so that I can tell exception causes apart), or leave this as is? I'd be glad with either.

zoriya: Let's leave it as is then, if it crashes due to a race condition it'll restart anyways.

PR #911, opened 2025-04-28
[v4, v5] Added support for storing transcoder metadata in S3

solidDoWant:
Similar to https://github.com/zoriya/Kyoo/pull/896, this adds support for storing transcode _metadata_ in S3. **This does not store transcoded video in S3, only files previously placed under `/metadata`**.

This required a number of changes:
* I bumped the Go version to 1.24. This is required to use [a new 1.24 security feature](https://go.dev/blog/osroot) that prevents directory traversal attacks. Tangentially related, the `SanitizePath` can and probably should switch to this.
* I completely replaced all filesystem operations on the `GOCODER_METADATA_ROOT` (`/metadata`) path with a backend-agnostic interface.
* I fixed a bug where thumbnails would not be (re)generated if the thumbnail sprite existed but the VTT file did not.
* I fixed a bug where subtitle and VTT files could be briefly unavailable if requested by another client while they were being regenerated.

One thing that this does not address is distributed locking. The RunLock struct only provides in-process locking. I'm a little unclear why this is needed with the metadata today, and so I'm not sure what the implications are of having the lock not be shared between multiple instances.

@zoriya if you don't mind giving this a once over, I'll add some unit tests should the overall approach look reasonable to you. I'm also going to merge this into my fork and run it for a bit to see what issues I hit.


transcoder/src/extract.go
```diff
@@ -15,48 +21,86 @@ func (s *MetadataService) ExtractSubs(info *MediaInfo) (interface{}, error) {
...
+	}
+
+	subtitlesWorkingDirectory := filepath.Join(tempWorkingDirectory, "sub")
+	if err := os.MkdirAll(subtitlesWorkingDirectory, 0660); err != nil {
+		return fmt.Errorf("failed to create subtitles directory: %w", err)
+	}
+
+	defer func() {
+		cleanupErr := os.RemoveAll(tempWorkingDirectory)
+		err = errors.Join(err, cleanupErr)
```
zoriya, line 92: `err` is assigned but can never be read. I think it's fine to just `os.RemoveAll()` without checking for the error.

transcoder/src/extract.go
```diff
@@ -68,16 +112,26 @@ func extractSubs(info *MediaInfo) error {
...
-	err := cmd.Run()
-	if err != nil {
+
+	if err := cmd.Run(); err != nil {
 		fmt.Println("Error starting ffmpeg extract:", err)
 		return err
 	}
+
+	// Move attachments and subtitles to the storage backend
+	// TODO should existing files be removed?
```
zoriya, line 128: the `defer os.RemoveAll(workdir)` should already handle that.

transcoder/src/storage/storage.go
```diff
@@ -0,0 +1,74 @@
...
+	// ListItemsWithPrefix returns a list of items in the storage backend that match the given prefix.
+	// Note: returned items may have a "/" in them, e.g. "foo/bar/baz".
+	ListItemsWithPrefix(ctx context.Context, pathPrefix string) ([]string, error)
+	// DeleteItem deletes an item from the storage backend. If the item does not exist, it returns nil.
+	// Deletion should be "syncronous" (i.e. the function should block until the write is complete).
+	DeleteItem(ctx context.Context, path string) error
+	// SaveItemWithCallback saves an item to the storage backend. If the item already exists, it overwrites it.
+	// The writeContents function is called with a writer to write the contents of the item.
+	// Writes should be "syncronous" (i.e. the function should block until the write is complete).
+	SaveItemWithCallback(ctx context.Context, path string, writeContents ContentsWriterCallback) error
```
zoriya, line 29: I think using a callback here makes it harder to reason about the code. Wouldn't this work:

```go
Save(ctx context.Context, path string, content io.Reader) error
```

`io.Reader` is the parameter of the parameter that takes aws & would work with an `io.Copy` for local fs.

transcoder/src/thumbnails.go
```diff
@@ -128,26 +157,73 @@ func extractThumbnail(path string, sha string) error {
...
-			os.Remove(f)
-		}
+	// Track which items existed before creating the new ones.
+	// It is impotant not to delete the items before writing the new ones,
+	// or requests for the old items will fail in the time between when the
+	// old items are deleted and the new ones are created.
+	oldItems, err := s.storage.ListItemsWithPrefix(ctx, getThumbPrefix(sha))
+	if err != nil {
+		return fmt.Errorf("failed to get old thumbnail items: %w", err)
 	}
```
zoriya, line 167: files could be removed upfront, this function holds a lock that prevent others from reading the file while it's processing.

(anyway this path only happens when we change the `ThumbsVersion` & want to delete all thumbnails generated with a prior version of the code)

transcoder/src/thumbnails.go
```diff
@@ -128,26 +157,73 @@ func extractThumbnail(path string, sha string) error {
...
+			if err := s.storage.DeleteItem(ctx, oldItem); err != nil {
+				return fmt.Errorf("failed to delete old thumbnail with path %q: %w", oldItem, err)
+			}
+			return nil
+		})
+	}
+	if err := deletionGroup.Wait(); err != nil {
+		return fmt.Errorf("failed to delete old thumbnails: %w", err)
+	}
+
```
zoriya, line 210: Instead of walking & deleting every file, I think we could put thumbs data in a subdirectory and delete the whole directory. This would remove the need for the `ListFileWithPrefix` function from the `storage` backend (but the Delete function would need to handle directories)

transcoder/src/storage/storage.go
```diff
@@ -0,0 +1,74 @@
...
+	// GetItem retrieves an item from the storage backend.
+	GetItem(ctx context.Context, path string) (io.ReadCloser, error)
+}
+
+// SaveFilesToBackend is a helper function that saves files from the source directory to the backend storage.
+func SaveFilesToBackend(ctx context.Context, backend StorageBackend, sourceDirectory, destinationBasePath string) error {
+	if err := filepath.WalkDir(sourceDirectory, func(sourcePath string, d fs.DirEntry, err error) error {
```
zoriya, line 36: I'm putting the comment here but this is true for other lines/files as well:

I think `if err := function(); err != nil` is a good pattern ONLY when `function()` is short. When `function()` is multiline I find it harder to read & i'd rather we have

```go
err := function(
	with,
	multiples,
	args,
)
if err != nil {}
```

transcoder/src/extract.go
solidDoWant: This is actually read. `err` here refers to the named return value, so both the error value from the "normal" logic and the cleanup error will be returned.

transcoder/src/extract.go
solidDoWant: I was more referring to the files in the storage backend. However, given the small file sizes, this is probably fine to leave until I implement some sort of cleanup for transcoded files. WDYT?

transcoder/src/thumbnails.go
solidDoWant: Is this desirable in all cases? For example:
* If saving the thumbnail files fail
* If image encoding fails
* If the process is restarted while in this section

If it is desirable to remove the contents in these cases, then I can simplify this as suggested. WDYT?

transcoder/src/thumbnails.go
solidDoWant: > I think we could put thumbs data in a subdirectory and delete the whole directory

They are all under a path prefix (not necessarily a directory) for a given sha currently.  See [here](https://github.com/zoriya/Kyoo/pull/911/files#diff-f62acd7327c32833a674bd63faa1a17f66df4d8088b272e73399258f05911d90L32-L34).

How about a separate `DeleteAllWithPrefix` function? This would allow different backends to perform deletions in the most optimal manner (`rm -rf` for file-based, `ListObjectsV2` + `DeleteObjects` for S3).

> This would remove the need for the ListFileWithPrefix

This logic will be needed with any approach that involves deleting files with a given path prefix with S3. However, with a `DeleteAllWithPrefix` function, I can remove this from the interface (for now) to maybe simplify it a bit.

transcoder/src/extract.go
zoriya: I think it's fine to never delete those files, we currently never delete them

transcoder/src/thumbnails.go
zoriya: oh you're right, didn't know s3 didn't have a delete (non empty) directory api.

I was thinking of something like `os.RemoveAll("/metadata/sha/thumb")` for fs.


Yeah I'm fine with either a `DeletePrefix` or just make the delete function do a list + delete if path ends with a /

transcoder/src/thumbnails.go
zoriya: Yes i think this is desirable, anyways we don't change thumbs code often so this case shouldn't be encountered frequently.

let's do with the simplest possible

transcoder/src/storage/storage.go
solidDoWant: I can do this, but the tradeoff is that anything that expects an io.writer (such as `imaging.Encode`) will need to re-implement reader/writer piping as well as concurrency (because the reader and writer need to be read/written to at the same time).

What would you think about two functions: one that takes a writer callback, and one that takes a simple io.Reader? Something like `SaveItemWithWriter` and SaveItemWithReader`, or `SaveItemWithCallback` and `SaveItem`?

transcoder/src/thumbnails.go
solidDoWant: > Yeah I'm fine with either a DeletePrefix or just make the delete function do a list + delete if path ends with a /

I think I'm going to add a separate function for this, and continue to have `DeleteItem` only delete a specific item. Some backends (s3) don't have a concept of directories (`/` is just another character in the object name), so I think it'd be better to make separate functions with explicit behavior, then add implementation-specific logic to all implementations of a single function.

transcoder/src/storage/storage.go
zoriya: That's a good point, didn't think about `imaging.Encode`. Having both `Save` & `SaveWithCallback` sounds good to me!

solidDoWant:
> We'd need to distribute the RunLock because we don't want all instances of the transcoder to race to extract subtitles/fonts/thumbs. (this is up to debate but i don't think we do).

Does this need to be a part of this PR? Or perhaps a better question, does the transcoder _currently_ support multiple instances? If not, then I'll split this out into another PR after this + the pgx PR merge.

transcoder/src/metadata.go
```diff
@@ -64,13 +119,31 @@ func NewMetadataService() (*MetadataService, error) {
...
-		lock:         NewRunLock[string, *MediaInfo](),
-		thumbLock:    NewRunLock[string, interface{}](),
-		extractLock:  NewRunLock[string, interface{}](),
-		keyframeLock: NewRunLock[KeyframeKey, *Keyframe](),
-	}, nil
+	return db, nil
+}
+
+func (s *MetadataService) setupStorage(ctx context.Context) (storage.StorageBackend, error) {
+	s3BucketName := os.Getenv("S3_BUCKET_NAME")
```
solidDoWant, line 126: Any opinion on whether or not I should add an explicit `BACKEND_TYPE` env var here? In the off chance that another backend is added at some point (such as some form of document storage), this could make the change + migrating significantly easier.

solidDoWant:
Here's an overview of things I changed in the last commit:
* Addressed all PR comments
* Fixed some leaked goroutines and file handles
* Changed how `main` handles errors so that they are handled consistently, including when deferred cleanup funcs are called
* Fixed a couple of errors that were not caught
* A couple of likely long-running sections that wait on external services are now ran concurrently via error groups
* I removed the "waiters" for S3 operations that wait until an operation is complete. Most common S3 backends are strongly consistent so this shouldn't be needed, and simplifies some operations significantly.

zoriya:
> Does this need to be a part of this PR?
no another PR would be better. Current transcoder doesn't support replication anyways

transcoder/main.go
```diff
@@ -217,11 +218,12 @@ func (h *Handler) GetAttachment(c echo.Context) error {
...
 	}
 
-	ret, err := h.metadata.GetAttachmentPath(sha, false, name)
+	ret, err := h.metadata.GetAttachment(sha, name)
+	defer ret.Close()
 	if err != nil {
 		return err
 	}
-	return c.File(ret)
+	return c.Stream(200, "TODO", ret)
```
zoriya, line 226: TODO that should be fixed before merging

transcoder/src/metadata.go
zoriya: both are fines to me

zoriya (approved):
Left minors comments but almost ready for merge!

transcoder/main.go
solidDoWant: any idea what the proper MIME types for these + attachments should be? I think they can both vary depending on what extension the underlying file uses.

solidDoWant:
Made a couple of changes:
* Replaced all TODO contexts. This required plumbing all requests contexts through to the new functions, so the diff is pretty simple but somewhat large. I did this on the pgx PR as well, but I'm also doing it here to ensure that I don't miss anything.
* I moved the utils file to its own package to make it available to the storage package without a circular dependency. Refs to it have been appropriately updated.
* I added a cleanup-with-error helper function. I then replaced all the `defer` calls I added with this function, to ensure that all errors stemming from deferred logic are properly captured. I checked all instance of `defer` in the codebase and now all these errors should be captured.
* I fixed another leaky goroutine bug with the S3 writer uploader. The changes simplified the function a bit by just using an error group, which is essentially what I had manually built before.

transcoder/main.go
```diff
@@ -290,19 +297,33 @@ type Handler struct {
 
 func main() {
 	e := echo.New()
+
+	if err := run(e); err != nil {
+		e.Logger.Fatal(err)
```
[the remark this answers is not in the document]
solidDoWant: How are stack traces actually being captured? Doesn't Go require that you capture the stack from when the error is created to preserve this info?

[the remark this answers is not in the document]
zoriya: mmm you are right actually, my bad

zoriya:
Do we actually care about `.Close` errors? I find the `utils.CleanupWithErr` with a named return & a ref to `&err` REALLY confusing.

solidDoWant:
Yea because Go requires explicit handling of errors this (or something effectively the same) is the only way to "bubble up" errors generated in deferred function calls

transcoder/src/metadata.go
solidDoWant: I'll leave it as is for now, so that it remains consistent with the other services

transcoder/main.go
solidDoWant: this should be fixed in the latest commit

solidDoWant:
@zoriya squashed and rebased. Should be good to go provided that I didn't make a git mistake

PR #913, opened 2025-04-29
Switch transcoder to pgx

solidDoWant:
As discussed on Discord, this switches the transcoder to use pgx. This is a simple lift and shift, with none of the new helper methods currently used.

The pgx library requires a context for most calls, so adding context args everywhere is a pretty big portion of this PR. **This has a very important design implication**. Pretty much all contexts stem from a given HTTP's request context, so when the request is cancelled (for example, when a client closes or drops the connection), the context will be cancelled. Because the context is passed to DB calls, cancellation of a client request may cause DB requests to return an error (even if they complete), potentially exposing bugs anywhere that errors are not handled properly.

I _think_ pgx natively supports queries with array parameters, as well as scanning into arrays, but I am not 100% sure. I've removed some PG type conversion for now under the assumption that this works properly with v5, but it will require some manual testing to verify.


transcoder/src/keyframes.go
```diff
@@ -70,25 +69,13 @@ func (kf *Keyframe) AddListener(callback func(keyframes []float64)) {
...
-	err := arr.Scan(src)
-	if err != nil {
-		return err
-	}
-	kf.Keyframes = arr
-	kf.IsDone = true
-	kf.info = &KeyframeInfo{}
-	return nil
-}
-
```
zoriya, line 84: This shouldn't be deleted! When we load from this from db (from `transcoder/src/metadata.go:191`), this won't set `IsDone` & won't know that only the array should be loaded from db.

solidDoWant: Yea I wasn't sure about this since it isn't explicitly referenced anywhere. Do you know how t his should be implemented with pgx? What type(s) will be provided as src?

transcoder/src/metadata.go
```diff
@@ -1,89 +1,134 @@
...
+	return &MetadataService{
+		database:     db,
+		lock:         NewRunLock[string, *MediaInfo](),
+		thumbLock:    NewRunLock[string, interface{}](),
+		extractLock:  NewRunLock[string, interface{}](),
+		keyframeLock: NewRunLock[KeyframeKey, *Keyframe](),
+	}, nil
+}
+
+func (s *MetadataService) doDBSetup(ctx context.Context) (*pgxpool.Pool, error) {
```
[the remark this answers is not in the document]
solidDoWant: I'll take care of this when I fix the merge conflicts that will occur when the S3 PR goes in

transcoder/src/keyframes.go
solidDoWant: As-is as in with the existing `Scan` func, or without it?

If you mean that the scan function should remain as-is, what type should I replace `pq.Float64Array` with (so that the dep can be dropped)?

zoriya:
Completed in #1135

PR #915, opened 2025-05-02
Added Go pprof handlers

solidDoWant:
Related discussion [here](https://discord.com/channels/1216460898139635753/1216460898609401870/1367367002607587361).

I've placed this in a new "api" package (though maybe "routers" would be a more apt name). IMO placing new routers and handlers in this package will help reduce `main.go` bloat over time as new endpoints are added. However, I'd be glad to move this into `main.go` if preferred.

It looks like by default the transcoder api binds to `/`, which can (and will) cause conflicts with all other non-API router groups (such as these endpoints). IMO the API should force an `/api` prefix to prevent this. Users could still add another prefix to this, but it would prevent accidently causing a route conflict, and would reduce the likelihood of exposing performance profile data.

### DIFF
--- a/transcoder/.env.example
+++ b/transcoder/.env.example
@@ -20,6 +20,12 @@ GOCODER_VAAPI_RENDERER="/dev/dri/renderD128"
 # the qsv device path (only used with GOCODER_HWACCEL=qsv)
 GOCODER_QSV_RENDERER="/dev/dri/renderD128"
 
+# Performance tuning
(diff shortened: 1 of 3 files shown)


zoriya (approved):

zoriya:
The `api`/`router` package is a good call.

This route is under the `/debug/pprof` so there are no chances of conflict, but yeah it could be a good design decision. Anyway, the default traefik config adds a prefix (`/api`, `/auth`, soon `/videos` & co). We might as well hard code them here instead of having the `_PREFIX` vars.

PR #931, opened 2025-05-05
Add support for QSV with newer Intel iGPUs

solidDoWant:
Newer Intel iGPUs require some additional packages for QSV transcoding. Tested this on an i9-13900H.

This works, but audio with QSV transcoding _specifically_ is about 1s behind. Not sure if this was introduced by these changes, or if there is another issue with QSV transcoding. I'd need somebody else to test on some older hardware to get a better idea of where the problem lies. VAAPI works fine on the same hardware.

### DIFF
--- a/transcoder/Dockerfile
+++ b/transcoder/Dockerfile
@@ -47,7 +47,9 @@ RUN set -x && apt-get update \
 	# hwaccel dependencies
 	vainfo mesa-va-drivers \
 	# intel hwaccel dependencies, not available everywhere
-	$([ " $TARGETARCH" = " amd64" ] && echo "intel-media-va-driver-non-free i965-va-driver-shaders") \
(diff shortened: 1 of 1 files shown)


solidDoWant:
Is there an easy way to check? Do I just download `0/<quality>/segment-<num>.ts` and `audio/0/segment-<num>.ts` for some video and plug them into ffprobe?

zoriya:
yes ffprobe or mediainfo are good tools for that. if you are running the `docker-compose.dev.yaml` file you can run it directly in the transcoder's root (`/tmp/kyoo_cache` by default), else you can download the segments via the api yeah.

zoriya (approved):
ping me when you think this is ready for merge, i can't test the audio desync (i don't own a qsv enabled igpu)

PR #942, opened 2025-05-06
Added support for considering DTS audio streams when determing player codec support

solidDoWant:
This adds the DTS codec type to the media MIME type, when a DTS stream is the primary audio stream.

This change will cause media info in the DB to be regenerated.

I'm not 100% sure if this is the correct value for the codec type (see link for values). FFmpeg reports the type as `DTS (dtc)`. Getting this wrong should be minimally impactful, as current clients utilizing this code path _very_ likely do not support any form of DTS audio streams.

### DIFF
--- a/transcoder/src/codec.go
+++ b/transcoder/src/codec.go
@@ -115,6 +115,11 @@ func GetMimeCodec(stream *ffprobe.Stream) *string {
 		ret := "alac"
 		return &ret
 
+	// Based on https://cconcolato.github.io/media-mime-support/#video/mp4;%20codecs=%22dtsc%22
+	case "dts":
+		ret := "dtsc"
+		return &ret
+
 	default:
 		log.Printf("No known mime format for: %s", stream.CodecName)
 		return nil
--- a/transcoder/src/codec.go
+++ b/transcoder/src/codec.go
@@ -115,9 +115,31 @@ func GetMimeCodec(stream *ffprobe.Stream) *string {
 		ret := "alac"
 		return &ret
 
-	// Based on https://cconcolato.github.io/media-mime-support/#video/mp4;%20codecs=%22dtsc%22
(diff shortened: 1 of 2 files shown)


transcoder/src/codec.go
```diff
@@ -115,6 +115,11 @@ func GetMimeCodec(stream *ffprobe.Stream) *string {
 		ret := "alac"
 		return &ret
 
+	// Based on https://cconcolato.github.io/media-mime-support/#video/mp4;%20codecs=%22dtsc%22
+	case "dts":
+		ret := "dtsc"
```
zoriya, line 120: I don't think your source is correct, example dts mime type computation from [exoplayer](https://github.com/google/ExoPlayer/blob/dd430f7053a1a3958deea3ead6a0565150c06bfc/library/common/src/main/java/com/google/android/exoplayer2/util/MimeTypes.java#L439-L446)

it looks like this is a more complicated mime format that require profiles & co embedded (ex dtsx is `/vnd.dts.uhd;profile=p2`)

solidDoWant: Is this good to go then or is there more action required on my part?

zoriya: i think you need to detect which variant of dts the file is in & set `ret` to either `dtsl`,`dtsx`, dtsh` & co accordingly

solidDoWant: Pushed a fix for this. Running a build now and will test to see if it works

zoriya (approved):

zoriya:
feel free to merge as you see fit

solidDoWant:
This seems to work. Example mime codec value for a video with DTS-HD MA audio: `"mimeCodec": "video/x-matroska; codecs=\"avc1.640029, dtsl\""`

> feel free to merge as you see fit

This should be good to go now, but I can't merge it myself (only contributors with write access can)

PR #944, opened 2025-05-07
Cleanly handle unsupported subtitles

solidDoWant:
This handles unsupported (primarily bitmap) embedded subtitles "nicely". On the backend, when only unsupported subtitle formats are embedded, ffmpeg is never called. This keeps the transcoder from logging an unimportant error (noise). On the frontend, these subtitles are now hidden in the player:

image

### DIFF
--- a/front/packages/ui/src/player/video.tsx
+++ b/front/packages/ui/src/player/video.tsx
@@ -109,12 +109,14 @@ const Video = forwardRef<VideoRef, VideoProps>(function Video(
 				}
 				// when video file is invalid, audio is undefined
 				selectedAudioTrack={{ type: SelectedTrackType.INDEX, value: audio?.index ?? 0 }}
-				textTracks={subtitles?.map((x) => ({
-					type: MimeTypes.get(x.codec) as any,
-					uri: x.link!,
-					title: x.title ?? "Unknown",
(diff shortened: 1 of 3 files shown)


front/packages/ui/src/player/video.tsx
```diff
@@ -109,7 +109,7 @@ const Video = forwardRef<VideoRef, VideoProps>(function Video(
 				}
 				// when video file is invalid, audio is undefined
 				selectedAudioTrack={{ type: SelectedTrackType.INDEX, value: audio?.index ?? 0 }}
-				textTracks={subtitles?.map((x) => ({
+				textTracks={subtitles?.filter(x => !!x.link).map((x) => ({
```
solidDoWant, line 112: I can't build, publish, or test this myself. However it follows identical logic to the web player, which I did test, so I think it'll work here.

zoriya (approved):

front/packages/ui/src/player/video.tsx
zoriya: yeah sounds good, exoplayer (current player on android) doesn't support pgs & co anyways

solidDoWant:
This should be good to go

PR #945, opened 2025-05-07
Reduce playback start time by 100x

solidDoWant:
Playback start time is dominated by the keyframe lookup process, which needs partial completion before the playback index can be generated and returned to the client. Currently, this requires a minimum of 100 keyframes, regardless of segment duration, to be parsed before this happens. This can take 2s to 2m, depending on segment size, file size, and disk IO load.

Rather than requiring a flat 100 keyframes, this PR changes the index generation logic to require both 5s worth of keyframes and a count of 3 keyframes. In many cases I am seeing a playback start time reduction of 100x or more, cutting this process down to about 100ms in most cases.

### DIFF
--- a/transcoder/src/keyframes.go
+++ b/transcoder/src/keyframes.go
@@ -26,13 +26,9 @@ type Keyframe struct {
 	info      *KeyframeInfo
 }
 type KeyframeInfo struct {
-	ready sync.WaitGroup
-	mutex sync.RWMutex
-	// This channel is closed when enough keyfraames for index generation
-	// have been parsed. A nil value also indicates that the keyframes have
-	// been parsed.
-	indexKeyframesParsed chan struct{}
-	listeners            []func(keyframes []float64)
+	ready     sync.WaitGroup
+	mutex     sync.RWMutex
+	listeners []func(keyframes []float64)
 }
 
 func (kf *Keyframe) Get(idx int32) float64 {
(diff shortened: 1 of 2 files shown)


transcoder/src/keyframes.go
```diff
@@ -14,17 +14,25 @@ import (
...
 type KeyframeInfo struct {
-	ready     sync.WaitGroup
-	mutex     sync.RWMutex
-	listeners []func(keyframes []float64)
+	ready sync.WaitGroup
+	mutex sync.RWMutex
+	// This channel is closed when enough keyfraames for index generation
+	// have been parsed. A nil value also indicates that the keyframes have
+	// been parsed.
+	indexKeyframesParsed chan struct{}
```
zoriya, line 34: this was already what the `ready sync.WaitGroup` was for!

transcoder/src/keyframes.go
```diff
@@ -231,6 +242,15 @@ func getVideoKeyframes(path string, video_idx uint32, kf *Keyframe) error {
...
 		ret = append(ret, fpts)
 
+		// Notify listeners that enough keyframes for index generation have been parsed
+		if !indexNotificationComplete && fpts >= minParsedKeyframeTime && len(ret) >= minParsedKeyframeCount {
+			kf.add(ret)
+			done += len(ret)
+			ret = ret[:0]
+			indexNotificationComplete = true
+			close(kf.info.indexKeyframesParsed)
+		}
```
zoriya, line 252: You can just update the condition below of
```go
if len(ret) == limit
```
to
 ```go
if (
	(!indexNotificationComplete && fpts >= minParsedKeyframeTime && len(ret) >= minParsedKeyframeCount)
	|| len(ret) == limit
) {
```

kf.add() already handles the wait group & notify the code in `stream.go`

zoriya (approved):
thanks, crazy speed gains!

Issue #566, opened 2024-07-14
Add an option to disable the regsiter button


zoriya:
### Feature description

We can ask for verification but not entierly disable the registration flow.

PR #947, opened 2025-05-08
Added support for disabling password login and registration

solidDoWant:
This adds support for disabling login via username/password, as well as registration. Enabling these for new installs requires that OIDC auth is configured.

Here's what the login page looks like with password auth disabled:

image

Here's what the registration page looks like with registration disabled:

image

Fixes https://github.com/zoriya/Kyoo/issues/566

### DIFF
--- a/front/packages/ui/src/login/login.tsx
+++ b/front/packages/ui/src/login/login.tsx
@@ -18,11 +18,10 @@
  * along with Kyoo. If not, see <https://www.gnu.org/licenses/>.
  */
 
-import { type QueryPage, login } from "@kyoo/models";
+import { type QueryPage, ServerInfoP, login, useFetch } from "@kyoo/models";
 import { A, Button, H1, Input, P, ts } from "@kyoo/primitives";
 import { useEffect, useState } from "react";
-import { useTranslation } from "react-i18next";
-import { Trans } from "react-i18next";
+import { Trans, useTranslation } from "react-i18next";
 import { Platform } from "react-native";
 import { useRouter } from "solito/router";
 import { percent, px, useYoshiki } from "yoshiki/native";
@@ -35,6 +34,11 @@ export const LoginPage: QueryPage<{ apiUrl?: string; error?: string }> = ({
 	apiUrl,
 	error: initialError,
 }) => {
+	const { data } = useFetch({
+		path: ["info"],
+		parser: ServerInfoP,
+	});
+
 	const [username, setUsername] = useState("");
 	const [password, setPassword] = useState("");
 	const [error, setError] = useState<string | undefined>(initialError);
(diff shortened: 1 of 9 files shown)


front/packages/ui/src/login/login.tsx
```diff
@@ -53,43 +57,47 @@ export const LoginPage: QueryPage<{ apiUrl?: string; error?: string }> = ({
...
-					m: ts(1),
-					width: px(250),
-					maxWidth: percent(100),
-					alignSelf: "center",
-					mY: ts(3),
-				})}
-			/>
+			<OidcLogin apiUrl={apiUrl} hideOr={!data?.passwordLoginEnabled} />
+			{data?.passwordLoginEnabled && (
+				<Array>
```
zoriya, line 62: woaw didn't even know `<Array>` was a thing (ig it works because it's just calling js's `new Array()` method).

Could you use fragments instead? (that's just `<>` & `</>`)

zoriya (approved):
Small comment but otherwise good to go.

For v5, you mentioned if it would be possible to simply disable (aka not start) the `auth` service & i've though about it recently. (i'm hacking this PR to talk about that lmao)

If we want to disable the `auth` service entirely, that would mean:
 - disable password login/registration (and only keep oauth flow, like in this PR)
 - disable admin pages about users (users permissions, verifications & co)

Those would be fine, but that means we lose the ability to have pages that display other users' history, stats & stuff like that (well we could have them, but we wouldn't be able to access their user info like name, pictures & so on, only the stats the api has & a user id/pk)

In v5, services will require a JWKS endpoint to validate the jwt they get as `Authorization: Bearer`. This is to validate the phantom token generated by the `auth` service, but it could as well be used to validate a token from `authentik` or any other oauth serv.
The downside of this approach would be: no phantom token & we only use the `access_token` from the oauth provider & the front won't try to refresh it.

front/packages/ui/src/login/login.tsx
solidDoWant: sure - done

solidDoWant:
@zoriya I should probably clarify a bit why I was asking about disabling the auth service. What I'm _actually_ interested in disabling/not using is phantom tokens, _especially_ if I need to host traefik just to handle exchanging them. I'm 100% fine with running another pod for user history, stats, settings, etc. (though it might be worth considering storing these per-service).

Basically, as a Kyoo user/admin, I like the current approach of logging in with my IdP, authenticating with Kyoo via OIDC. I know that there are more complexities when there isn't a single service "proxying" all requests like there is today, but I think that these can be solved without adding Traefik to exchange tokens in every case.

What do you think about this (poorly drawn) approach:
```
+--------+             +--------------------+                            +---------------------+
|        |             |    API gateway     |                            |         Auth        |
| Client | <----+--->  |                    | <-------------+----------> |  +---------------+  |
|        |      |      |  If token missing, |               |            |  | Password auth |  |
+--------+      |      |  redirect to auth  |               |            |  +---------------+  |
    ^           |      |                    |               |            |                     |
    |           |      +--------------------+               |            |  +---------------+  |
    v           |                                           |            |  |   Anon auth   |  |
+--------+      |                                           |            |  +---------------+  |
|        |      |                                           |            |                     |
|  IdP   | <----+                                           |            |  +---------------+  |
|        |                                                  |            |  | JWKS provider |
+--------+                                                  |            |  +---------------+  |
                                                            |            |                     |
                                                            |            +---------------------+
                                                            |                       ^
                                                            |                       |
                                                            |                       v
                                                            |            +---------------------+
                                                            +----------> |     Some service    |
                                                                         +---------------------+
```

Here's how requests would work:

* When a token header is missing, the gateway would send a 401 unauthorized response.
  * Specific routes for login would not return a 401.
  * Clients would parse the parse the response and redirect to a login page.
* When a service receives a request with a token, it verifies the signature.
  * Services will periodically check the auth service for a new set of keys. Should a message broker be re-added, they could just listen for "broadcast" type messages for when a new set is available.
  * If the signature is expired, send a 401 unauthorized response. The client can contact then handle refreshing it (ideally silently via a refresh token).
* When a user wants to log in, clients can request a list of available login methods from the auth service.
* When using password auth, the auth service "password auth" section would essentially act as a mini-IdP. It would store common user info (like name), and would generate a JWT and refresh token as external IdPs would.
* For anonymous access (if enabled), clients would silently "log in" and receive a valid "guest" JWT from the "anon auth" section of the auth service.
* Communication between services on behalf of a user would use the user's token. This ensures that when service "A" makes a request to service "B" on behalf of a user, service "B" can validate (via JWT claims) that the actual user has permission to perform the API call.
* In the future, if support is needed so that services can make API calls on behalf of a user at a later time (where the original user token may be expired), services could request a tightly-scoped, but long-lived JWT from the auth service to perform actions on the user's behalf. I don't think that anything needs this today though.
* Service to service auth that is _not_ on behalf of a user would be handled via an unsigned JWTs.
  * Securing this (if desired) is the responsibility of the Kyoo admin. There's a few ways to do this that I can discuss if this is a sticking point.
  * To differentiate between a user-provided, unsigned JWT, and a service-provided, signed JWT, service JWTs would use a different header.
  * The API gateway would be required to drop requests with this header set (or at least drop the header).
  * Alternatively, using a service could entirely communicate via a message queue. This is probably a more scalable approach, and would solve some other problems (like sending notifications/messages to all instances of a service).
  * Alternatively, mTLS could be used. This is pretty easy to setup, as long as cert issuance and renewal is handled outside of kyoo via cert manager/similar.
* Each service be responsible for storing and retrieving user-specific data (settings, stats, etc.), along with RBAC. The JWT should have enough information to support this.

Disclaimer: I'm pretty on OIDC so I may have overlook something or be relying on something infeasible.

TL; DR I think this is achievable without traefik or phantom tokens. Use OIDC or an internal password auth JWT source to generate a JWT, and have every service handle user-specific data separately. Leave service identity attestation to the admin.

btw if you're interested in taking this discussion further, it might make sense to move it to a separate issue or discord so that other folks can chime in.

zoriya:
Move to [discord](https://discord.com/channels/1216460898139635753/1216460898609401870/1370482613827010702). We might wanna try using gh discussions or something better for search later on idk

PR #1025, opened 2025-07-23
Transcoder: If empty JWKS env var, do not enable JWKS

Arthi-chaud:


transcoder/main.go
```diff
@@ -84,32 +84,34 @@ func main() {
...
-			var pubkey interface{}
-			if err := jwk.Export(key, &pubkey); err != nil {
-				return nil, fmt.Errorf("Unable to get the public key. Error: %s", err.Error())
-			}
-
-			return pubkey, nil
-		},
-	}))
+
+	if src.Settings.JwksUrl != "" {
```
zoriya, line 89: This `if` could start l82 before the `jwk.NewCache()`

zoriya (approved):
thanks!

PR #1133, opened 2025-11-01
fix(api): don't crash if not superuser

Intuinewin:
While testing kyoo v5, I was unable to launch the api because of this query which requires superuser rights (owner of the db is not enough):
`ALTER DATABASE "${postgresConfig.database}" SET pg_trgm.word_similarity_threshold = 0.4;`

I believe Kyoo should be usable without superuser rights so I added a try/catch as a quickfix to only display an error on startup.

### DIFF
--- a/api/src/db/index.ts
+++ b/api/src/db/index.ts
@@ -116,13 +116,17 @@ export const db = drizzle({
 });
 
 export const migrate = async () => {
-	await db.execute(
-		sql.raw(`
+	try {
+		await db.execute(
+			sql.raw(`
 			create extension if not exists pg_trgm;
 			SET pg_trgm.word_similarity_threshold = 0.4;
 			ALTER DATABASE "${postgresConfig.database}" SET pg_trgm.word_similarity_threshold = 0.4;
 		`),
-	);
+		);
+	} catch (err: any) {
+		console.error("Error while updating pg_trgm", err.message);
+	}
 	await migrateDb(db, {
 		migrationsSchema: "kyoo",
 		migrationsFolder: "./drizzle",


acelinkio:
Definitely would like to have a discussion about the expectations Kyoo with Postgres.  

Ideally Kyoo would not require any super administrator permissions.  Can be the owner of the database it connects to, but should be limited to not be a DBMS administrator.

zoriya:
Agreed on the change, we should just have a documentation section to specify we need the trigram extension of postgres.

acelinkio:
Terminology around databases has a lot of overloaded terms.  Going to list out a few terms to be explicit.

| FullName                  | Shorthand | Description                                  |
|---------------------------|-----------|----------------------------------------------|
| DatabaseManagementSystem  | DBMS      | Postgres Instance                            |
| Database                  | DB        | Database inside of the DBMS                  |
| PostgresSchema            | Schema    | Postgres Schema inside of the DB             |
| Tables/Sequences/Etc      | n/a       | Objects that live inside of Postgres' schema |

I'm all for shift left, where developers are responsible for everything and allow their applications to do anything.  The way of limiting their blast radius is giving each application their own DBMS--where they also become responsible for backups, monitoring, pooling, etc...

However if they can adhere to a few constraints, they can use a shared DBMS that's already maintained.  That constraint is doing everything as a DB owner.  DB owners can do pretty much anything inside of that database... They can create schemas, manage indexes, and even drop tables.  However they cannot install extensions.  Extensions are installed at the DBMS level to make it available to the instance and then installed into schemas.  That installation requires super admin and also the schema has to already exist.

So there can be a bit of back and forth between who should be responsible for schemas.  Developers need to create the schema then operators need to come back and install the extension.

I argue that schema creation/management should be on the operator.  But the application/developer is responsible for describing everything they need upfront.  That schema can be created when the database is initialized and extensions installed before an application starts. In 95% of cases, only one schema is needed and that one schema name should not matter. So that's why I would prefer to just rely upon the user's search_path setting--as it is database specific.

acelinkio:
To add to what this looks like from a development perspective.  You can still have a single user setup and use multiple databases with ease.  

Have scripts to create database/schema.  **Example:  create the `kyoo_api` database with the `uuid-ossp` extension**

```sql
CREATE DATABASE kyoo_api WITH OWNER kyoo_all; 
\connect kyoo_api;
REVOKE ALL ON SCHEMA public FROM PUBLIC;
CREATE SCHEMA IF NOT EXISTS data AUTHORIZATION kyoo_all;
CREATE EXTENSION IF NOT EXISTS "uuid-ossp" SCHEMA data;
```

Configure the user's search_path.  search_path as it is specific to each database.  **Example: configure the `kyoo_all` user's `search_path` setting.** 
```sql
ALTER ROLE kyoo_all
IN DATABASE kyoo_auth SET search_path TO "$user", data;
ALTER ROLE kyoo_all 
IN DATABASE kyoo_api SET search_path TO "$user", data;
ALTER ROLE kyoo_all 
IN DATABASE kyoo_scanner SET search_path TO "$user", data;
ALTER ROLE kyoo_all 
IN DATABASE kyoo_transcoder SET search_path TO "$user", data;
```

zoriya:
search_path will NOT be configurable (some orm i use simply don't support that). Most (if not all) queries/statement will use fully qualified names (like `select * from kyoo.entries`). each service has their own schema, they can all use the same db/dbms or each have their own. both use cases are supported

for now, only the trigram extension is needed. i think the `try/catch` approach is the good one in the code but we should just add a bit of a doc to explain to users which extension/settings are needed.

PR #1134, opened 2025-11-02
fix(front): use auth token to load images on web

Intuinewin:
I'm not sure if the comment `// use cookies on web to allow img to make the call instead of js` is still relevant

### DIFF
--- a/front/src/models/video-info.ts
+++ b/front/src/models/video-info.ts
@@ -47,7 +47,7 @@ export const Subtitle = z.object({
 	language: z.string().nullable(),
 	codec: z.string(),
 	mimeCodec: z.string().nullable(),
-	extension: z.string(),
+	extension: z.string().nullable(),
 	isDefault: z.boolean(),
 	isForced: z.boolean(),
 	isHearingImpaired: z.boolean(),


zoriya:
Yeah that comment is still relevent, if we want to use headers instead of cookies for images/videos in the web, we need to download images in js and convert them to base64 blobs.

This is a huge hit on performance compared to just letting the browser do the request.

The auth service should automatically handle cookies as well as headers for authorization so this should be transparent for services.

Why do you need that?

Intuinewin:
> The auth service should automatically handle cookies as well as headers for authorization so this should be transparent for services.

Is it supposed to work already?

When trying to load an image the api answers with `{"status":403,"message":"Missing permission: 'core.read'.","details":{"current":[],"required":["core.read"]}}`. I see the cookie in request headers but in the auth service I can't find any code to handle the cookie.

In my setup: `GUEST_CLAIMS={"permissions": [], "verified": false}`

zoriya:
Mmm I'll check that when i get back home

zoriya:
you're right, i've fixed it in #1135

Intuinewin:
> you're right, i've fixed it in https://github.com/zoriya/Kyoo/pull/1135

Thank you 🚀  I've reverted the condition change but I kept this PR open as it still contains small front fixes

Issue #968, opened 2025-05-28
Road to v5


zoriya:
This is the list of things to do before we can release v5. Priorities are between https://github.com/zoriya/Kyoo/labels/P0 and https://github.com/zoriya/Kyoo/labels/P3:
 - https://github.com/zoriya/Kyoo/labels/P0 is a must-have for kyoo to run (so requirements for first alpha version)
 - https://github.com/zoriya/Kyoo/labels/P1 is a blocker for final v5 release. Probably a breaking change we need to tackle before we stabilize the api/database.
 - https://github.com/zoriya/Kyoo/labels/P2/https://github.com/zoriya/Kyoo/labels/P3 are nice to have, not blocking for the final v5 release but most of them would have been implemented before

Front:
 - [ ] https://github.com/zoriya/Kyoo/labels/P0 Migrate from next+expo to a single project (either one with #763 or maybe just expo if one isn't ready)
 - [ ] https://github.com/zoriya/Kyoo/labels/P0 Update everything for new APIs
 - [ ] https://github.com/zoriya/Kyoo/labels/P1 Reimplement SSR if #763 disabled it
 - [ ] https://github.com/zoriya/Kyoo/labels/P1 Virtualize lists on the web (maybe legends-list? #270)
 - [ ] https://github.com/zoriya/Kyoo/labels/P2 Redesign logo/colors
 - [ ] https://github.com/zoriya/Kyoo/labels/P3 complete visual update
 - [ ] https://github.com/zoriya/Kyoo/labels/P3 switch to tailwind (#268)
 - [ ] https://github.com/zoriya/Kyoo/labels/P3 display staff list
 - [ ] https://github.com/zoriya/Kyoo/labels/P3 Add history page
 - [ ] https://github.com/zoriya/Kyoo/labels/P3 Add user stats page

Scanner:
 - [ ] https://github.com/zoriya/Kyoo/labels/P0 finish & merge #923 
 - [ ] https://github.com/zoriya/Kyoo/labels/P1 Implement & use tvdb for all series by default
 - [ ] https://github.com/zoriya/Kyoo/labels/P1 Implement support for AnimeList #466 
 - [ ] https://github.com/zoriya/Kyoo/labels/P1 If AnimeList support is not enough, implement back thexem support
 - [ ] https://github.com/zoriya/Kyoo/labels/P1 Implement back metadata refresh
[...]

PR #1141, opened 2025-11-05
feat(auth): update forward auth endpoint

Intuinewin:
Why this change?

In my tests with Envoy, the [ext_authz](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto) filter used for external authorization doesn't directly call the `/auth/jwt` endpoint, but adds a prefix to the queried path, resulting in a middleware call like this: `GET /auth/jwt/api/movies`. Furthermore, the HTTP method is preserved.

The GEP-1494, mentionned by @acelinkio in https://github.com/zoriya/Kyoo/issues/968#issuecomment-3482350025, established an API based on the behavior of the ext_authz filter (https://gateway-api.sigs.k8s.io/geps/gep-1494/#why-envoys-ext_authz) so IMO Kyoo should handle this behavior.

I also added Hurl tests to test cookie handling.

### DIFF
--- a/auth/main.go
+++ b/auth/main.go
@@ -19,10 +19,10 @@ import (
 	_ "github.com/golang-migrate/migrate/v4/source/file"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/jackc/pgx/v5/stdlib"
-	"github.com/labstack/echo-jwt/v4"
+	echojwt "github.com/labstack/echo-jwt/v4"
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
-	"github.com/swaggo/echo-swagger"
+	echoSwagger "github.com/swaggo/echo-swagger"
 )
 
 func ErrorHandler(err error, c echo.Context) {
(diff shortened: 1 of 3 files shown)


zoriya:
Is there a reason the `/auth/jwt` method needs to be changed? couldn't the current method (with it's token in the return body) be used for external auth if we listen to `c.ANY("/auth/jwt")` and `c.ANY("/auth/jwt/*")` ?

Intuinewin:
> Is there a reason the /auth/jwt method needs to be changed? couldn't the current method (with it's token in the return body) be used for external auth if we listen to c.ANY("/auth/jwt") and c.ANY("/auth/jwt/*") ?

Yes, you're right, the current method can be used. Since the JWT in the request body isn't necessary for external authentication, I simply thought it would be clearer to have a dedicated route. I can revert this change and use the same route if you prefer

zoriya:
let's keep a single http method (`/api/jwt`) then, id rather have fewer than two methods that do mostly the same thing.

thanks for the PR & the explanations of why, always appreciated.

PR #1173, opened 2025-11-21
add devspace

acelinkio:
DevSpace is a CNCF project designed to enable fast, iterative remote development directly inside Kubernetes. It shares a similar goal with Telepresence—improving the inner-loop development cycle for cloud-native applications—but approaches the problem differently.

Where Telepresence proxies your local machine’s network into the cluster, DevSpace instead has you develop inside the target container environment. Under the hood, DevSpace temporarily scales your existing deployment down to zero and replaces it with a development pod built from your chosen dev image. It then syncs your local source files into the pod and runs your defined startup command, giving you a tight edit-build-run loop with the same dependencies and environment as production.

For Kyoo, DevSpace effectively consolidates what was previously split across Dockerfile.dev and docker-compose.dev.yml. The development image, commands, and overrides you would normally maintain in local Docker tooling are now cleanly layered on top of your existing Kubernetes manifests, reducing duplication.  (duplication like in docker-compose.yml and docker-compose.dev.yml)

DevSpace can be extended to replace more of the `docker` functionality including image builds/tag/push  https://www.devspace.sh/docs/configuration/images/  

This PR introduces DevSpace for:
* api - leverages Bun's hot reload
* auth - uses DevSpace to stop/rerun command
* front - leverages Bun's hot reload
* scanner - uses DevSpace to stop/rerun command
* transcoder - uses DevSpace to stop/rerun command

Test it out:
```sh
nix-shell -p devspace
devspace use context
# set your kube config to default to the desired namespace
devspace use namespace kyoo-next
devspace dev
# when done
devspace purge
```
it even hosts a webserver that provides a quick view of the namespace and provides easy way to look at logs/exec/explore
https://localhost:8090 
image

Reference:  https://www.devspace.sh/docs/getting-started/development

### DIFF
--- a/api/devspace.yaml
+++ b/api/devspace.yaml
@@ -14,7 +14,7 @@ dev:
           exec:
             - command: bun install --frozen-lockfile
               onChange: 
-               - "./bun.lock"
+                - "./bun.lock"
     command: 
       - bash 
       - -c 
--- a/front/devspace.yaml
+++ b/front/devspace.yaml
@@ -10,6 +10,11 @@ dev:
         excludePaths:
           - node_modules
         startContainer: true
+        onUpload:
+          exec:
+            - command: bun install --frozen-lockfile
+              onChange: 
+                - "./bun.lock"
     # increased sysctl limits for file watching
     # front uses Metro javascript bundler which watches a lot of files
     # these are node level settings that should be raised


scanner/devspace.yaml
```diff
@@ -0,0 +1,27 @@
...
+        startContainer: true
+        onUpload:
+          restartContainer: true
+    command:
+      - bash
+      - -c
+      - |
+        echo "installing uv..."
+        pip install uv
+        echo "Running uv sync..."
```
zoriya: can't we just use a uv docker image?

api/shell.nix
```diff
@@ -10,6 +10,7 @@ pkgs.mkShell {
     node-gyp
     pkg-config
     vips
+    devspace
```
zoriya: Can we put this under the root shell.nix istead of each one?

zoriya:
thank you !

PR #1174, opened 2025-11-21
Allow manual helm chart publishing for specific tags

joryirving:
This lets us cut `:master` helm charts for publishing/OCI, to enable easier testing.
Also updates the actions as well as pins then to digests for security purposes.


joryirving:
Honestly, fair point. I was thinking if you wanted to publish older/previous tags, but I don't think that's needed.

PR #1183, opened 2025-11-25
kyoo_api extension install specify schema 

acelinkio:
explicitly specify schema for extension install rather than rely upon implicit

https://www.postgresql.org/docs/current/sql-createextension.html
> schema_name
    The name of the schema in which to install the extension's objects, given that the extension allows its contents to be relocated. The named schema must already exist. If not specified, and the extension's control file does not specify a schema either, the current default object creation schema is used.

### DIFF
--- a/api/src/db/index.ts
+++ b/api/src/db/index.ts
@@ -128,8 +128,6 @@ export const migrate = record("migrate", async () => {
 		await db.execute(
 			sql.raw(`
 				create schema if not exists ${APP_SCHEMA};
-				alter database "${postgresConfig.database}" set search_path = ${APP_SCHEMA}, public;
-				set search_path = ${APP_SCHEMA}, public;
 				create extension if not exists pg_trgm schema ${APP_SCHEMA};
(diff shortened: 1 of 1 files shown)


acelinkio:
Prior to this change, tp_trgm was being installed into the public schema for any fresh rdbms/database.  The extension installation runs before any migrations scripts are ran and the `kyoo` schema is not created yet.

When explicitly setting where to install the postgres extension, we need that schema to exist and do that by adding:  `create extension if not exists pg_trgm schema kyoo;`.

```sql
SELECT opcnamespace::regnamespace, opcname
FROM pg_opclass
WHERE opcname = 'gin_trgm_ops' or opcname = 'gist_trgm_ops';
```

should return
| opcnamespace::regnamespace | opcname       |
|----------------------------|---------------|
| kyoo                       | gist_trgm_ops |
| kyoo                       | gin_trgm_ops  |


Things still work when using the `public` schema but lets try avoiding it where ever possible.

zoriya:
Thanks! just rebased to fix merge conflicts.

PR #1186, opened 2025-11-28
Fix pg admin password

zoriya:
Cloudlnative pg did a breaking change on the PATCH release 0.12.2 to fix this: https://github.com/CloudPirates-io/helm-charts/issues/619

### DIFF
--- a/chart/Chart.lock
+++ b/chart/Chart.lock
@@ -1,0 +1,6 @@
+dependencies:
+- name: postgres
+  repository: oci://registry-1.docker.io/cloudpirates
+  version: 0.12.4
+digest: sha256:e486b44703c7a97eee25f7715ab040d197d79c41ea1c422ae009b1f68985f544
+generated: "2025-12-01T20:17:25.152279487+01:00"
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -12,4 +12,4 @@ dependencies:
 - condition: postgres.enabled
   name: postgres
   repository: oci://registry-1.docker.io/cloudpirates
-  version: 0.12.0
+  version: 0.12.4


acelinkio:
Didn't mean to remove draft.  Can't readd on mobile

zoriya:
dw, idk what i did wrong but it's refusing to render on my argo :c

> Unable to save changes: application spec for kyoo-v5 is invalid: InvalidSpecError: Unable to generate manifests in chart: rpc error: code = Unknown desc = failed to execute helm template command: failed to get command args to log: `helm template . --name-template kyoo-v5 --namespace kyoo-next --kube-version 1.33 --values /tmp/227bf4e5-8b0a-4f03-b1b4-4b103fd02b2f <api versions removed> --include-crds` failed exit status 1: Error: YAML parse error on kyoo/charts/postgres/templates/statefulset.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{".Values.global.postgres.infra.passwordKey":interface {}(nil)} Use --debug flag to render out invalid YAML

zoriya:
Turns out they also changed their logic and now it doesn't re-render the values: https://github.com/CloudPirates-io/helm-charts/commit/c0c832a5ffecf155471ea1f4eeb2288c81107dea?w=1

I don't think we can inline the passwordKey from another Value.

acelinkio:
Going to lookk through what they did.  This is how I originally added some of it

https://github.com/CloudPirates-io/helm-charts/pull/513

acelinkio:
Opened upstream pull request.  https://github.com/CloudPirates-io/helm-charts/pull/674

also added this chart to my own argocd here: https://github.com/acelinkio/argocd-homelab/blob/main/pineapp/kyoo-dev/manifest.yaml

zoriya:
tryed it in my cluster and still got another random error :c
image

acelinkio:
they added a breaking change  can't mutate that field.  https://github.com/CloudPirates-io/helm-charts/pull/681

image


if you delete the statefulset, argocd will recreate it with the labels and use the same PVC

zoriya:
so they are doing breaking changes in every patch release? is that expected?

zoriya:
should we just pin a digest and never update it?

acelinkio:
Bitnami was the most reliable/safest before the rug pull.  They largely took care of the upgrade paths between even major versions. CloudPirates is trying to fill the void and appear to be going through some learning pains. 

I'm not opposed to just pinning a digest and updating once every few months. Users should not rely upon subcharts for anything stateful. Subcharts for any resources are really only good for quick start or development purposes.   I tried to emphasize that in the readme.

https://github.com/zoriya/Kyoo/tree/master/chart#subchart-support
> Subcharts are updated frequently and subject to changes. This chart includes subcharts for deploying PostgreSQL. Please consider hosting those independently of Kyoo to better handle versioning and lifecycle management.

Using a postgres operator is definitely the way to host postgres because they handle all of the advanced configuration including things like pgpool, leader election, high availability, upgrades, wal backups, extensions, users, multi-writer/sharding(?), etc.  wal backs will always be better than backing up PV/PVCs.

SideRant:
Honestly the less PV/PVCs we create/manage the better.  ReadOnlyMany for accessing your media library is 100% fine. ReadWriteMany is to share files between pods is not reliable.  Not all CSIs support ReadWriteMany and some that do use NFS under the hood which has its own reliability concerns.  

At the end of the day, distributed storage is hard.  It's a reason why I am hoping to eventually move S3 for storage.  It moves state into an external systems, handles multiple writes, offers object lifecycle management.

zoriya:
yeah i agree, and i use a pg operator too but i feel like the straightforward path should just work and not break every 3d. if it's learning pain i 100% understand and that's fine but i hope they stop doing so many breaking changes.

PR #1194, opened 2025-12-01
scanner: refactor otel integration

acelinkio:
Refactoring the OTEL integration to make what handles what a bit cleaner.  OTEL standards handles metrics, spans/traces, and logs.  Logging still requires an external logging library.  https://opentelemetry.io/docs/specs/otel/logs/ 

OTEL introduces double logging where you are logging to stdout and logging directly to the OTEL connector.   Taking that approach avoids needing to parse/process or force stdout logs into a particular format or enriched.  STDOUT can remain human readable.

This PR introduces a:
* EnvVariable `STDOUT_LOG_LEVEL` which provides a way to configure what is written to stdout.  For those that are using OTEL, they can set this to warn/error and reduce the amount of double logging (sending to stdout and to OTEL)
* Does not connect to OTEL unless `OTEL_EXPORTER_OTLP_ENDPOINT` is specified,  Which prevents those reoccurring transport errors that exists that happen right now
* Simplifies `OTEL_EXPORTER_OTLP_PROTOCOL` based lib imports
* Breaks logging and otel into two different modules.  OTEL configuration can be done entirely via https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration

### DIFF
--- a/scanner/scanner/otel.py
+++ b/scanner/scanner/otel.py
@@ -26,39 +26,26 @@ def setup_otelproviders() -> tuple[object, object, object]:
 
 	# choose exporters (grpc vs http) ...
 	if os.getenv("OTEL_EXPORTER_OTLP_PROTOCOL", "").lower().strip() == "grpc":
-		try:
-			from opentelemetry.exporter.otlp.proto.grpc._log_exporter import (
-				OTLPLogExporter,
-			)
-			from opentelemetry.exporter.otlp.proto.grpc.metric_exporter import (
-				OTLPMetricExporter,
-			)
-			from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
-				OTLPSpanExporter,
-			)
-		except Exception as exc:
-			raise RuntimeError(
-				"gRPC OTLP exporter imports failed. Install the necessary packages / system libs."
-			) from exc
-		else:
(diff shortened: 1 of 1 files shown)


acelinkio:
Created a sample project here https://github.com/acelinkio/python-otel-example where I experimented.  

Would like to aim to have the other api/auth/front/transcoder eventually use `STDOUT_LOG_LEVEL` env variable.  Eventually it would be nice to have all of kyoo have the same stdout logging format/style.

scanner/scanner/otel.py
```diff
@@ -1,80 +1,91 @@
...
+			logger.info("Using HTTP libs for OpenTelemetry exporter.")
+
+	from opentelemetry import trace, metrics, _logs
+	from opentelemetry.sdk.trace import TracerProvider
+	from opentelemetry.sdk.trace.export import BatchSpanProcessor
+	from opentelemetry.sdk.metrics import MeterProvider
+	from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
+	from opentelemetry.sdk._logs import LoggerProvider
+	from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
+	from opentelemetry.sdk.resources import Resource
```
zoriya: why move imports here?

PR #1197, opened 2025-12-02
scanner: adding the probes back

acelinkio:
https://github.com/zoriya/Kyoo/pull/1195 moved the probes routes.  adding them back to avoid crashloops


acelinkio:
@zoriya small code fix.  self-merging

PR #1204, opened 2025-12-07
auth: slog, otel, & logging improvements

acelinkio:
similar to https://github.com/zoriya/Kyoo/pull/1194

This pull request does the following:
* Refactors otel.go
* Use otel noop when OTEL_EXPORTER_OTLP_ENDPOINT is empty
* Add slog as the logger
* Sends echo logs through slog
* Send slogs to both stdout and otel
* Adds echo filter to avoid logging paths.  ~Controllable via LOG_IGNORE_WEBPATHS env var~
* Add loglevel filter.  Controllable via STDOUT_LOG_LEVEL and  OTEL_LOG_LEVEL env vars

stdout logs before change
```log
Migrating database
Migrating finished

   ____    __
  / __/___/ /  ___
 / _// __/ _ \/ _ \
/___/\__/_//_/\___/ v4.13.4
High performance, minimalist Go web framework
https://echo.labstack.com
____________________________________O/_______
                                    O\
⇨ http server started on [::]:4568
2025/12/07 00:57:25 failed to upload metrics: Post "https://localhost:4318/v1/metrics": dial tcp [::1]:4318: connect: connection refused
{"time":"2025-12-07T00:57:47.450738389Z","id":"","remote_ip":"10.42.3.190","host":"kyoo-dev-auth:4568","method":"GET","uri":"/.well-known/jwks.json","user_agent":"Go-http-client/1.1","status":200,"error":"","latency":469325,"latency_human":"469.325µs","bytes_in":0,"bytes_out":811}
```

stdout logs after change
```log
level=INFO msg="Configuring OTEL"
level=INFO msg="Using OLTP type" type=noop
level=INFO msg="Skipping request logging for these paths" paths=/.well-known/jwks.json,/health,/ready
level=INFO msg="Database migration state" state=starting
level=INFO msg="Database migration state" state=completed
level=INFO msg="⇨ http server started on [::]:4568"
```

logs do ship to the opentelemetry-collector now!

### DIFF
--- a/auth/main.go
+++ b/auth/main.go
@@ -251,8 +251,8 @@ func main() {
 
 	ignorepath := []string{
 		"/.well-known/jwks.json",
-		"/health",
-		"/ready",
+		"/auth/health",
+		"/auth/ready",
 	}
 	slog.Info("Skipping request logging for these paths", "paths", func() string { sort.Strings(ignorepath); return strings.Join(ignorepath, ",") }())
 


acelinkio:
Could consider generalizing `otel.go` and `slog.go` so they can be copy/pasted into transcoder project or published into an library.

auth/main.go
```diff
@@ -227,17 +229,96 @@ func (h *Handler) TokenToJwt(next echo.HandlerFunc) echo.HandlerFunc {
...
+			for _, p := range strings.Split(val, ",") {
+				if p = strings.TrimSpace(p); p != "" {
+					ignore = append(ignore, p)
+				}
+			}
+		}
+	}
+	sort.Strings(ignore)
+	stringignore := strings.Join(ignore, ",")
+	slog.Info("web_request.log_ignore_paths", "paths", stringignore)
```
zoriya: why do you make this configurable? this feels uneeded to me.

also favicon.ico doesn't need to be part of that list, it wouldn't be routed to this service.

auth/main.go
```diff
@@ -227,17 +229,96 @@ func (h *Handler) TokenToJwt(next echo.HandlerFunc) echo.HandlerFunc {
...
+
+	contains := func(list []string, s string) bool {
+		for _, v := range list {
+			if v == s {
+				return true
+			}
+		}
+		return false
+	}
```
zoriya: there's `slices.Contains` now. you don't need to write it yourself.

zoriya:
i think we could make this into a library quite easily since all git repositories are libraries in golang.

we would just need to place those files in a directory inside the auth directory and add an import into the transcoder like `import "github.com/zoriya/kyoo/auth/logs"`

auth/main.go
acelinkio: I think what I'll try to do move these to log at debug.

acelinkio:
Awesome to hear you are open to creating internal libraries!  Can look at moving to a library approach in the future.

PR #1219, opened 2025-12-14
purge unused

acelinkio:
MR for removing unused code.  Clean up can be done when ever, just trying to take care of some of the tedium


zoriya:
I didn't delete those because i want to `git move` them when reimplementing them. I work a lot with git history, so i don't wanna lose it.

We can delete the `autosync` module tho, the code will be too different for a git move to make sense anyway.

acelinkio:
I'll open a seperate MR for deleting autosync. I'm not quite ready for git surgery, moving things around.

PR #1243, opened 2025-12-27
fix: add cache scope to prevent multi-platform build collisions

fenio:
Add scope parameter to GHA cache configuration to ensure each image (api, front, scanner, transcoder, auth) gets its own cache namespace. This prevents cache collisions that were causing arm64 builds to fail for kyoo_auth images.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -219,6 +219,9 @@ api:
   extraContainers: []
   extraInitContainers: []
   extraVolumes: []
+  # resourceClaims for Dynamic Resource Allocation (DRA)
+  # useful for accessing hardware resources via ResourceClaimTemplates (Kubernetes 1.34+)
+  resourceClaims: []
   # api image data
   # user profile pictures
   persistence:
@@ -269,6 +272,9 @@ auth:
   extraContainers: []
   extraInitContainers: []
   extraVolumes: []
+  # resourceClaims for Dynamic Resource Allocation (DRA)
+  # useful for accessing hardware resources via ResourceClaimTemplates (Kubernetes 1.34+)
+  resourceClaims: []
 
 # front deployment configuration
 front:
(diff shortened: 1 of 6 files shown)


acelinkio:
@joryirving 

Iirc you helped out with the GitHub actions and also suggested the ResourceTemplates 

Could I get your help on reviewing this?

chart/values.yaml
```diff
@@ -219,6 +219,9 @@ api:
   extraContainers: []
   extraInitContainers: []
   extraVolumes: []
+  # resourceClaims for Dynamic Resource Allocation (DRA)
+  # useful for accessing hardware resources via ResourceClaimTemplates (Kubernetes 1.34+)
+  resourceClaims: []
```
joryirving: I'm pretty sure we only need this for the transcoder, and nothing else.

joryirving (commented):
A few comments. I think we only need resourceClaim on the transcoder, and I think we can extrapolate the resource for it using the resource claim. Feel free to clean up the suggested code, or ditch it completely.

acelinkio:
I don't quite know enough about resource claims atm.  Should we just consume claims and assume they are already there or is it something this helm chart should manage the lifecycle of?

joryirving:
You need a resourceClaimTemplate (which is managed outside the helm chart).
You make a claim (I do this inside the bjw-s app-template), and you specify the resource it creates.

Examples:
ResourceclaimTemplate: https://github.com/joryirving/home-ops/blob/32a70e6f6a8aa53eb9df2afbbb2c93381017dd26/kubernetes/components/gpu/resourceclaimtemplate.yaml
ResourceClaim: https://github.com/joryirving/home-ops/blob/32a70e6f6a8aa53eb9df2afbbb2c93381017dd26/kubernetes/apps/base/llm/ollama/helmrelease.yaml#L61-L63
Resource: https://github.com/joryirving/home-ops/blob/32a70e6f6a8aa53eb9df2afbbb2c93381017dd26/kubernetes/apps/base/llm/ollama/helmrelease.yaml#L42-L44

acelinkio:
Thanks for the info.  Fenio mentioned in discord they were going to recreate these MRs.

PR #1246, opened 2025-12-28
fix: add cache scope to prevent multi-platform build collisions

fenio:
Add scope parameter to GHA cache configuration to ensure each image (api, front, scanner, transcoder, auth) gets its own cache namespace. This prevents cache collisions that were causing arm64 builds to fail for kyoo_auth images.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -107,8 +107,8 @@ jobs:
             VERSION=0.0.0
           push: ${{env.SHOULD_PUSH}}
           tags: ${{steps.meta.outputs.tags}}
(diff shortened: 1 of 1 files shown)


acelinkio:
I'm kind of confused how it broke in the first place.  I'm not as familiar with Github actions as I should be to know if this will fix it.

fenio:
@acelinkio I think this is one of those cases where it worked well because we were lucky enough, and the problem had always been there.

And, after all, what's the harm in checking if this option will help? It's just a scope change that makes the key unique. If it doesn't help, you can always roll it back later.

PR #1263, opened 2026-01-10
api: fix tracing

acelinkio:
Adding tp.register() fixes the missing methods / path from showing.

image

i thought that setGloabl calls would handle registering.  the javascript otel docs are a bit lacking--esepcially compared to golang.
```ts
	logapi.setGlobalLoggerProvider(lp);
	metricapi.setGlobalMeterProvider(mp);
	tp.register();
	traceapi.setGlobalTracerProvider(tp);
```

### DIFF
--- a/api/src/otel.ts
+++ b/api/src/otel.ts
@@ -124,7 +124,6 @@ export function setupOtel() {
 	logapi.setGlobalLoggerProvider(lp);
 	metricapi.setGlobalMeterProvider(mp);
 	tp.register();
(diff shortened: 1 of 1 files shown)


acelinkio:
tp.register() calls setGlobalTracerProvider() 
https://github.com/open-telemetry/opentelemetry-js/blob/v2.3.0/packages/opentelemetry-sdk-trace-node/src/NodeTracerProvider.ts#L89-L101

image


however I can't find similar commands for lp/mp.  This appears to be specific to tp.

PR #1324, opened 2026-02-21
auth: echo v5 update

acelinkio:


acelinkio:
Majority of the upgrade to v5 is complete.  I don't quite understand what is happening with the auth tests failing.

One of the tests says it should be returning a 201, but gets a 409 conflict. The other shows claims are not being returned.  I don't quite know what got jumbled in the mix here. 

Will continue investigating, greatly appreciate any thoughts you may have @zoriya

zoriya:
i think your issue in the test file is that the `change-password` test is aborted because you get a 500 so the test account is never deleted. the same account seem to be created again in another test, if the first test failed (and the account didn't get deleted), you'll get a duplicated error.

acelinkio:
Fixed the error by adding post via `g.POST("/jwt", h.CreateJwt)`.  Seems like v5 doesn't catch it from `g.Any("/jwt/*", h.CreateJwt)`

Ready for review.  Following feedback I can do Transcoder as well.

auth/main.go
```diff
@@ -356,11 +355,20 @@ func main() {
 	r.DELETE("/keys/:id", h.DeleteApiKey)
 
 	g.GET("/jwt", h.CreateJwt)
+	g.POST("/jwt", h.CreateJwt)
```
zoriya:
```suggestion
g.Any("/jwt", h.CreateJwt)
```

PR #1486, opened 2026-04-29
Fix audio discontinuity

zoriya:
Since #1353, we allow transcoding of audios in the original quality and there's a discontinuity at head boundaries that makes player skip/replay a few ms.

This is an attempt to fix this issue, the hls_tests are slop useful for debugging this specific issue (debugging this by hand is pure hell). It's in no way a proper automated test suite.

Current state of the branch is maybe a bit better, maybe not i'm not sure. It definitely need more work and i'm not even sure the approach of editing keyframe detection for audio is the right way.

The easy fix would be to simply disable `-c:a copy` and always transcode but that's just sad.


zoriya:
Using output seeking for those files seems to work, i'll merge that and monitor in my cluster.

PR #1576, opened 2026-07-11
Whitelist oidc redirect urls

zoriya:
Only allow redirection after oidc to whitelisted urls (or first-party clients)


zoriya:
I  can't test this right now (i'm moving and my auth server is offline). Will merge as soon as possible

PR #1620, opened 2026-08-01
Transcoder: Fix parsing of keyframes for audio

Arthi-chaud:
The first line output by ffprobe us suffixed with a coma (just like for videos). However, this coma gets in the way of processing the keyframe's timestamp.

This PR remove the trailing coma, fixing transcoding of audio files


zoriya (approved):
weird that you need this for some files but lgtm

